PR #58, opened 2019-02-11
update gardenctl logs command

KristianZH:
With this PR the `gardenctl logs` command will be able to take the logs both from `kubectl` and `elasticsearch`. There are several new flags we can use in the command:
`--elasticsearch`(For getting the logs from elasticsearch)
`--tail=X`(Getting the last X logs)
`--since=Xs/m/h`(Getting the logs from last X seconds/minutes/hours)
`--since-time=RFC3339-format`(Getting the logs from after the specified time)

CC: @ialidzhikov @vlvasilev @vlerenc @vpnachev @plkokanov

### DIFF
--- a/cmd/logs.go
+++ b/cmd/logs.go
@@ -15,9 +15,11 @@
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


cmd/logs.go
```diff
@@ -140,16 +185,63 @@ func logPod(toMatch string, toTarget string, container string) {
...
+	if container != "" {
+		container = " -c " + container
+	}
 	pods, err := Client.CoreV1().Pods(namespace).List(metav1.ListOptions{})
 	checkError(err)
 	for _, pod := range pods.Items {
 		if strings.Contains(pod.Name, toMatch) {
-			err := ExecCmd(nil, "kubectl logs --tail="+numberOfLines+" "+pod.Name+container+" -n "+namespace, false, "KUBECONFIG="+KUBECONFIG)
+			err := ExecCmd(nil, buildKubectlCommand(namespace, pod.Name, container), false, "KUBECONFIG="+KUBECONFIG)
+			fmt.Println(buildKubectlCommand(namespace, pod.Name, container))
```
vlvasilev: this must run only for debugging purpose.

cmd/logs.go
```diff
@@ -140,16 +185,63 @@ func logPod(toMatch string, toTarget string, container string) {
...
 			checkError(err)
 		}
 	}
 }
 
+func showLogsFromElasticsearch(namespace, toMatch, container string) {
+	commands := buildElasticsearchCommands(namespace, toMatch, container)
+	fmt.Println()
```
vlvasilev: what is this for?

cmd/logs.go
```diff
@@ -140,16 +185,63 @@ func logPod(toMatch string, toTarget string, container string) {
...
+}
+
+func buildKubectlCommand(namespace, podName, container string) string {
+	var command strings.Builder
+	command.WriteString(fmt.Sprintf("kubectl logs %s%s -n %s ", podName, container, namespace))
+	if flags.tail != -1 {
+		command.WriteString(fmt.Sprintf("--tail=%d ", flags.tail))
+	}
+	if flags.sinceSeconds != -0 {
+		command.WriteString(fmt.Sprintf("--since=%ds ", flags.sinceSeconds/1000000000))
```
vlvasilev: Can we just use flags.sinceSeconds.Seconds()?

cmd/logs.go
```diff
@@ -371,3 +463,158 @@ func logsDNS() {
...
+	sb.WriteString("]}")
+	return sb.String()
+}
+
+func newRequestQuery(from, size int, r rng, sources []string, terms []term) *requestQuery {
+	q := new(requestQuery)
+	q.BoolQuery.rng = r
+	q.BoolQuery.term = append(q.BoolQuery.term, terms...)
+	q.From = int64(from)
+	q.Size = int64(size)
```
vlvasilev: q.Size can't exceed 10000. Why is int64

cmd/logs.go
```diff
@@ -26,106 +28,149 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	maxEsLogsPerQuery     = 10000
+	fourteenDaysInSeconds = 60 * 60 * 24 * 14
+)
+
 // numberOfLines of logfile to return
 var numberOfLines = "400"
```
ialidzhikov: @DockToFuture , @KristianZH what is the purpose of this? Do we still need it as now we have a `--tail` param?

cmd/logs.go
```diff
@@ -371,3 +461,158 @@ func logsDNS() {
...
+				Timestamp time.Time `json:"@timestamp"`
+				Log       string    `json:"log"`
+				Severity  string    `json:"severity"`
+				Source    string    `json:"source"`
+			} `json:"_source"`
+		} `json:"hits"`
+	} `json:"hits"`
+}
+
+type rng struct {
```
ialidzhikov: You could give a better name. Maybe `range` is a better name (if rng means range in your context)?

cmd/logs.go
```diff
@@ -140,16 +185,61 @@ func logPod(toMatch string, toTarget string, container string) {
...
+			err := ExecCmd(nil, buildKubectlCommand(namespace, pod.Name, container), false, "KUBECONFIG="+KUBECONFIG)
 			checkError(err)
 		}
 	}
 }
 
+func showLogsFromElasticsearch(namespace, toMatch, container string) {
+	commands := buildElasticsearchCommands(namespace, toMatch, container)
+	for _, command := range commands {
+		output, err := ExecCmdReturnOutput("bash", "-c", fmt.Sprintf("export KUBECONFIG=%s; %s", KUBECONFIG, command))
```
ialidzhikov: Isn't possible to pass the path to kubeconfig with the `--kubeconfig` param of kubectl?

cmd/logs.go
```diff
@@ -26,106 +28,149 @@ import (
...
+			validateArgs(args)
+			validateFlags(flags)
+		},
+		Run: func(cmd *cobra.Command, args []string) {
+			runCommand(args)
+		},
+		ValidArgs: []string{"gardener-apiserver", "gardener-controller-manager", "ui", "api", "scheduler", "controller-manager", "etcd-operator", "etcd-main", "etcd-events", "addon-manager", "vpn-seed", "vpn-shoot", "auto-node-repair", "dashboard", "prometheus", "grafana", "alertmanager", "tf"},
+		Aliases:   []string{"log"},
+	}
+	cmd.Flags().Int64Var(&flags.tail, "tail", flags.tail, "Lines of recent log file to display. Defaults to 100 with no selector, showing all log lines otherwise 10, if a selector is provided.")
```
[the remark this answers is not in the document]
vpnachev: Generally, I do not like the default value to be the default value of the data type. So I recommend the default values of the flags to be explicitly set.

cmd/logs.go
```diff
@@ -26,106 +28,149 @@ import (
...
+	cmd.Flags().DurationVar(&flags.sinceSeconds, "since", flags.sinceSeconds, "Only return logs newer than a relative duration like 5s, 2m, or 3h. Defaults to all logs. Only one of since-time / since may be used.")
+	cmd.Flags().StringVar(&flags.sinceTime, "since-time", flags.sinceTime, "Only return logs after a specific date (RFC3339). Defaults to all logs. Only one of since-time / since may be used.")
+	cmd.Flags().BoolVar(&flags.elasticsearch, "elasticsearch", flags.elasticsearch, "Show logs from elasticsearch")
```
vpnachev: The help does not state what is happening when the flag is not set.
`Show logs from elasticsearch` -> `If the flag is set the logs are retrieved from the elasticsearch, otherwise from the kubelet`, for example.

cmd/logs.go
```diff
@@ -26,106 +28,149 @@ import (
...
+	return cmd
+}
+
+func validateArgs(args []string) {
+	if len(args) < 1 || len(args) > 2 {
+		fmt.Println("Command must be in the format: logs (gardener-apiserver|gardener-controller-manager|ui|api|scheduler|controller-manager|etcd-operator|etcd-main[etcd backup-restore]|etcd-events[etcd backup-restore]|addon-manager|vpn-seed|vpn-shoot|machine-controller-manager|dashboard|prometheus|grafana|alertmanager|tf (infra|dns|ingress) args(--elasticsearch|--tail|--since|--since-time|--timestamps)")
```
vpnachev: `args(--elasticsearch|--tail|--since|--since-time|--timestamps)` these are flags, not args.

cmd/logs.go
```diff
@@ -371,3 +461,158 @@ func logsDNS() {
...
+	}
+	return queries
+}
+
+func (response logResponse) String() string {
+	//var output strings.Builder
+	output := make([]string, 0)
+	for _, hit := range response.Hits.Hits {
+		source := hit.Source
+		output = append(output, fmt.Sprintf("%v %v %v | %v", source.Timestamp, source.Severity, source.Source, source.Log))
```
vpnachev: If you want the output to be aligned, you could check https://golang.org/pkg/text/tabwriter/

cmd/logs.go
```diff
@@ -26,106 +28,149 @@ import (
...
-			fmt.Println("Command must be in the format: logs (gardener-apiserver|gardener-controller-manager|ui|api|scheduler|controller-manager|etcd-operator|etcd-main[etcd backup-restore]|etcd-events[etcd backup-restore]|addon-manager|vpn-seed|vpn-shoot|machine-controller-manager|dashboard|prometheus|grafana|alertmanager|tf (infra|dns|ingress)")
+//flags passed to the command
+var flags *logFlags
+
+func newCmdLogs() *cobra.Command {
+	flags = newLogsFlags()
+	cmd := &cobra.Command{
+		Use:   "logs (gardener-apiserver|gardener-controller-manager|ui|api|scheduler|controller-manager|etcd-operator|etcd-main[etcd backup-restore]|etcd-main-backup|etcd-events[etcd backup-restore]|addon-manager|vpn-seed|vpn-shoot|machine-controller-manager|dashboard|prometheus|grafana|alertmanager|tf (infra|dns|ingress)",
+		Short: "Show and optionally follow logs of given component\n",
+		Long:  `s`,
```
vpnachev: `Long` is the long description(https://godoc.org/github.com/spf13/cobra#Command) of the command, so I do not get why it is set to `s` ?
/cc @DockToFuture

cmd/logs.go
DockToFuture: range is a keyword of golang and therefore not usable

DockToFuture:
@KristianZH, @vpnachev thank your for your feedback fixed it.

cmd/logs.go
KristianZH: Yup, but the problem is that `range` is reserved keyword in Golang

cmd/logs.go
```diff
@@ -26,106 +28,149 @@ import (
...
-			os.Exit(2)
-		} else if len(t.Target) == 0 {
-			fmt.Println("Target stack is empty")
-			os.Exit(2)
+	}
+}
+
+func runCommand(args []string) {
+	switch args[0] {
+	case "gardener-apiserver":
```
ialidzhikov: We could also add `--elasticsearch` to support gardener-apiserver, gardener-controller-manager.

cmd/logs.go
```diff
@@ -26,106 +28,149 @@ import (
...
+		logsGardenerApiserver()
+	case "gardener-controller-manager":
+		logsGardenerControllerManager()
+	case "ui":
+		logsUI()
```
ialidzhikov: What is the idea of this one?

```golang
// logsUI
func logsUI() {
	logPodGarden("gardener", "garden")
}
```

I guess we mean gardener-dashboard, right? Also we have:
```
gardenctl logs dashboard
gardenctl logs ui
```

Which one of them is the k8s dashboard and which one the gardener one? Can we use better naming?

cmd/logs.go
```diff
@@ -26,106 +28,149 @@ import (
...
+		},
+		Run: func(cmd *cobra.Command, args []string) {
+			runCommand(args)
+		},
+		ValidArgs: []string{"gardener-apiserver", "gardener-controller-manager", "ui", "api", "scheduler", "controller-manager", "etcd-operator", "etcd-main", "etcd-events", "addon-manager", "vpn-seed", "vpn-shoot", "auto-node-repair", "dashboard", "prometheus", "grafana", "alertmanager", "tf"},
+		Aliases:   []string{"log"},
+	}
+	cmd.Flags().Int64Var(&flags.tail, "tail", flags.tail, "Lines of recent log file to display. Defaults to 100 with no selector, showing all log lines otherwise 10, if a selector is provided.")
+	cmd.Flags().DurationVar(&flags.sinceSeconds, "since", flags.sinceSeconds, "Only return logs newer than a relative duration like 5s, 2m, or 3h. Defaults to all logs. Only one of since-time / since may be used.")
+	cmd.Flags().StringVar(&flags.sinceTime, "since-time", flags.sinceTime, "Only return logs after a specific date (RFC3339). Defaults to all logs. Only one of since-time / since may be used.")
```
ialidzhikov: since and since-time are ignored for components like gardener-apiserver, gardener-controller-manager, ui.

cmd/logs.go
ialidzhikov: Ignore this one, sorry, my mistake.

cmd/logs.go
DockToFuture: Dashboard is the Kubernetes dashboard and UI is ours.

cmd/logs.go
```diff
@@ -371,3 +461,158 @@ func logsDNS() {
...
+		}
+		queries = append(queries, *newRequestQuery(from, size, r, sources, terms))
+		from += (maxEsLogsPerQuery + 1)
+	}
+	return queries
+}
+
+func (response logResponse) String() string {
+	//var output strings.Builder
+	output := make([]string, 0)
```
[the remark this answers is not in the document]
DockToFuture: You are right, we should show the newest logs at the end to be persistent.

cmd/logs.go
DockToFuture: We could rename it.

PR #59, opened 2019-02-12
Update README.md

ialidzhikov:
Update gardenctl version in README.md to the lates one.

Fixes #

**Release note**:

```improvement operator
NONE
```


DockToFuture:
I would rather add a comment to the installation manual.
For example, to download version v0.1.0  or 0.7.0 on macOS, type:
```curl -LO https://github.com/gardener/gardenctl/releases/download/0.1.0/gardenctl-darwin-amd64``` instead of updating the manual on every release or we should provide a version file with the latest stable version like provided in kubectl and update it on every release in the concurse pipeline.
``` curl -LO https://github.com/gardener/gardenctl/releases/download/$(https://github.com/gardener/gardenctl/blob/master/stable-version)gardenctl-darwin-amd64```

PR #75, opened 2019-04-13
Refactor for testability

ialidzhikov:
Refactor for testability.

Fixes #

**Release note**:

```improvement operator
Refactor for testability, introducing gomock, adding tests.
```


DockToFuture:
Is there a reason why you've decided to choose ```counterfeiter``` over ```gomock```?
I would rather stick with the last one, similar to gardener.

ialidzhikov:
Nope, there was no special reason. Switched to `gomock`. It's even better now, the generated code is much less and it is possible to expect with concrete params - the `counterfeiter` way was to expect at the end that your fake was called with the appropriate params.

ialidzhikov:
`doc.go` introduced.

Issue #83, opened 2019-06-19
gardenctl {shell,ssh} does not appear to work when OS is JeOS


marwinski:
**Describe the bug**
```gardenctl shell``` and ```gardenctl ssh``` don't work when the operating system on the nodes is JeOS.

**To Reproduce**
```
$ gardenctl version
gardenctl:
	version     : 0.10.0
	build date  : 2019-05-02
	go version  : go1.12.4
	go compiler : gc
	platform    : darwin/amd64
$ gardenctl shell
ip-10-250-2-103.eu-central-1.compute.internal
$ gardenctl shell ip-10-250-2-103.eu-central-1.compute.internal
Error: node "ip-10-250-2-103.eu-central-1.compute.internal" not found
$ gardenctl ssh
Node ips:
- 10.250.2.103
$ gardenctl ssh 10.250.2.103
Downloaded id_rsa key
Creating bastion host
usage: aws [options] <command> <subcommand> [<subcommand> ...] [parameters]
To see help text, you can run:

  aws help
  aws <command> help
  aws <command> <subcommand> help
aws: error: argument --image-id: expected one argument
$
```

cluster is dev - core - dm-jeos3

**Expected behavior**
It should work

**Screenshots**
n/a

**Gardenctl Version (please complete the following information):**

see above

**Additional context**

ialidzhikov:
Thanks for reporting! #84 proposed for the shell cmd.

PR #84, opened 2019-06-20
Improve nodeSelector in shell cmd

ialidzhikov:
The value of kubernetes.io/hostname may be the same as the Node name in some environments and a different value in other environments ([source](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#interlude-built-in-node-labels)).
The PR removes the `strings.Contains(host["kubernetes.io/hostname"], name)` check which was preventing shell cmd to work with JeOS nodes.

Ref #83 

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/cmd/drop_test.go
+++ b/cmd/drop_test.go
@@ -37,7 +37,7 @@ var _ = Describe("Drop command", func() {
 			return command.Execute()
 		}
 	)
-	
+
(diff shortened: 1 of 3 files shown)


ialidzhikov:
Just updated the PR. Now should be fine for reviews. :)

PR #92, opened 2019-07-05
Prefix tags with 'v'

ialidzhikov:
Ref https://github.com/gardener/gardener/issues/1096#issuecomment-508649092

Fixes #

**Release note**:

```improvement operator
The release tags from now are prefixed with `v`.
```


Makefile
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 DATE=$(shell date -u  +%Y-%m-%d)
-VERSION=$(shell cat VERSION | sed 's/[-dev]//g')
+VERSION=$(shell cat VERSION)
```
DockToFuture: Now also the ```-dev```suffix would be added, is this intended?

ialidzhikov: I think that cicd calls the .ci/build with the appropriate version and `make build` is for local builds.

@ccwienk

DockToFuture: Let's stick with the suffix removal for now.

ialidzhikov: Done.

PR #94, opened 2019-07-10
Adapt gardenctl to use project resource instead of ns with project label

### LINKED ISSUE #89: Update gardenctl to use project resource for ls and target cmd
Use ```Project``` resource in gardenctl for listing and targeting projects instead of namespaces with ```garden.sapcloud.io/role=project``` label.

DockToFuture:
Adapt gardenctl to use project resource instead of namespace with project label for target and ls cmd.
Name of project is now correctly shown in target stack instead of namespace name.
Fixes #89 

**Release note**:

```improvement operator
Project names are now shown in target stack.
```


cmd/ls.go
```diff
@@ -106,21 +106,18 @@ func NewLsCmd(reader ConfigReader, ioStreams IOStreams) *cobra.Command {
 
 // getProjectsWithShoots lists list of projects with shoots
 func getProjectsWithShoots() {
-	projectLabel := "garden.sapcloud.io/role=project"
-	projectList, err := Client.CoreV1().Namespaces().List(metav1.ListOptions{
-		LabelSelector: fmt.Sprintf("%s", projectLabel),
-	})
-	checkError(err)
 	gardenClientset, err := clientset.NewForConfig(NewConfigFromBytes(*kubeconfig))
```
ialidzhikov:
```suggestion
	gardenClientset, err := target.GardenerClient()
```

cmd/var.go
```diff
@@ -20,28 +20,33 @@ import (
...
-var password string
-
-// file pathes
+const (
+	// configuration for gardenctl
+	suggestionsMinimumDistance int  = 2
+	prefixMatching             bool = true
+
+	// NamespaceForProject contains the namespace of a project
+	namespaceForProject string = "project.garden.sapcloud.io/namespace"
```
ialidzhikov: This label/annotation is removed with gardener/gardener#1035. Can't we use `.spec.namespace`?

cmd/target.go
```diff
@@ -232,14 +226,9 @@ func NewTargetCmd(targetReader TargetReader, targetWriter TargetWriter, configRe
...
-	tmp := KUBECONFIG
-
-	client, err := target.K8SClientToKind(TargetKindGarden)
+	gardenClientset, err := target.GardenerClient()
 	checkError(err)
-	projectLabel := "garden.sapcloud.io/role=project"
-	projectList, err := client.CoreV1().Namespaces().List(metav1.ListOptions{
-		LabelSelector: projectLabel,
-	})
+	projectList, err := gardenClientset.GardenV1beta1().Projects().List(metav1.ListOptions{})
```
ialidzhikov: I think it would be better whether we have `*` in the project - if that is the case we can list them all. But if it is just a exact name, we could directly get it by name.

ialidzhikov:
Can you check `make test`?

cmd/target.go
```diff
@@ -259,12 +254,8 @@ func resolveNameProject(target TargetInterface, name string) (matches []string)
 				matches = append(matches, project.Name)
 			}
 		} else {
```
ialidzhikov: You could remove the empty else.

cmd/target.go
```diff
@@ -232,14 +226,15 @@ func NewTargetCmd(targetReader TargetReader, targetWriter TargetWriter, configRe
...
-	client, err := target.K8SClientToKind(TargetKindGarden)
+	gardenClientset, err := target.GardenerClient()
 	checkError(err)
-	projectLabel := "garden.sapcloud.io/role=project"
-	projectList, err := client.CoreV1().Namespaces().List(metav1.ListOptions{
-		LabelSelector: projectLabel,
-	})
+	if !strings.Contains(name, "*") {
+		project, err := gardenClientset.GardenV1beta1().Projects().Get(name, metav1.GetOptions{})
+		checkError(err)
```
ialidzhikov:
```suggestion
if err != nil {
	return []string{}
}
```
so we can preserve also the unit test.

cmd/target_test.go
```diff
@@ -89,25 +88,6 @@ var _ = Describe("Target command", func() {
 	})
 
 	Context("with garden target", func() {
-		It("targeting project with wrong name", func() {
```
ialidzhikov: You could preserve the test.

PR #111, opened 2019-08-14
Improve completion

### LINKED ISSUE #102: Gardenctl completion
`gardenctl completion` should not write to a file but to stdout.

Also, we should follow the `kubectl` design:
`gardenctl completion bash`
and
`gardenctl completion zsh` <-- new feature :-)

vpnachev:
Output the completion script to the std out.
Add experimental support for zsh.

Fixes #102

/cc @vasu1124 

**Release note**:

```action user
`gardenctl completion` now accepts shell subcommand, `bash` or `zsh`.
```
```action user
`gardenctl completion <shell>` no longer print the completion script to the file `gardenctl_completion.sh` file, but on the standard output.
```
```improvement user
`gardenctl completion` has an experimental support for `zsh` shell.
```

### DIFF
--- a/pkg/cmd/completion.go
+++ b/pkg/cmd/completion.go
@@ -15,24 +15,39 @@
 package cmd
 
 import (
+	"os"
+
 	"github.com/spf13/cobra"
 )
 
 // completionCmd represents the completion command
 var completionCmd = &cobra.Command{
-	Use:   "completion",
-	Short: "Generate bash-completion file\n",
+	Use:   "completion <bash|zsh>",
+	Short: "Generate bash or zsh completion script\n",
+	Long:  ``,
+}
+
+var bashCompletionCmd = &cobra.Command{
+	Use:   "bash",
+	Short: "Generate bash completion script",
 	Long:  ``,
 	Run: func(cmd *cobra.Command, args []string) {
-		generateBashCompletion()
+		err := RootCmd.GenBashCompletion(os.Stdout)
+		checkError(err)
+	},
+}
+
+var zshCompletionCmd = &cobra.Command{
+	Use:   "zsh",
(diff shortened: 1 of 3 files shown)


pkg/cmd/completion.go
```diff
@@ -15,24 +15,39 @@
...
+		err := RootCmd.GenBashCompletion(os.Stdout)
+		checkError(err)
 	},
 }
 
-func init() {
+var zshCompletionCmd = &cobra.Command{
+	Use:   "zsh",
+	Short: "Generate zsh completion script",
+	Long:  ``,
```
ialidzhikov: This is not needed.

pkg/cmd/completion.go
```diff
@@ -15,24 +15,39 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
```
ialidzhikov: You could prefer the usage of RunE and return err. `checkError` and `os.Exit` is something that break the testability.

pkg/cmd/completion.go
```diff
@@ -15,24 +15,39 @@
...
+		err := RootCmd.GenZshCompletion(os.Stdout)
+		checkError(err)
```
ialidzhikov:
```suggestion
		return err
```

pkg/cmd/completion.go
```diff
@@ -15,24 +15,39 @@
...
+	},
 }
 
-//generate bash-completion file for gardenctl
-func generateBashCompletion() {
-	err := RootCmd.GenBashCompletionFile("gardenctl_completion.sh")
-	checkError(err)
+func init() {
```
ialidzhikov: You could move that to `root.go` to have only in single place all the cmds.

PR #112, opened 2019-08-15
Unified ImageID for creating the bastion

KristianZH:
Unified ImageID for creating the bastion. This will help us for determining the `ssh` command.
The image is fetched from the ControllerRegistration of the Cloud Provider.
Soon there will be a new user for the nodes, called `gardener`. After it we will simply ssh with bastion user `core` and node user `gardener`

Fixes #

@ialidzhikov  @DockToFuture @vpnachev 

-->
```improvement operator
Bastion host is created with `coreos` imageID.
```


pkg/cmd/ssh.go
```diff
@@ -100,6 +114,7 @@ func NewSSHCmd(reader TargetReader, ioStreams IOStreams) *cobra.Command {
 			switch infraType {
 			case "aws":
 				var imageID string
```
DockToFuture: lets move those 4 lines into the sshToAWSNode function

pkg/cmd/ssh.go
```diff
@@ -159,7 +173,7 @@ func sshToAWSNode(imageID, nodeIP, path string) {
 			break
 		}
 	}
-	bastionNode := "<user>@" + ip
+	bastionNode := "core@" + ip
 	node := "<user>@" + nodeIP
 	fmt.Println("- Fill in the placeholders and run the following command to ssh onto the target node. For more information about the user, you can check the documentation of the cloud provider:")
```
DockToFuture: I guess we no longer need this comment when using coreos ami image. Can you it.

pkg/cmd/ssh.go
```diff
@@ -279,22 +293,64 @@ func getNodeNameForIP(ip string) string {
...
+			}
+		}
+	}
+
+	if ami == "" {
+		fmt.Println(fmt.Sprintf("Cannot find AMI with coreos image for region: %s", nodeRegion))
 	}
 
-	return output, nil
+	return ami, nil
```
DockToFuture: if there is no case were an error is returned we can remove the error return value

pkg/cmd/ssh_for_alicloud.go
```diff
@@ -584,3 +592,42 @@ func (a *AliyunInstanceAttribute) getMinimumInstanceSpec() string {
...
+	err = json.Unmarshal(controllerRegistration.Spec.Deployment.ProviderConfig.Raw, &controllerRegistrationSpec)
+	checkError(err)
+	machineImages := controllerRegistrationSpec.Values.Config.MachineImages
+	imageID := machineImages[0].ID
+
+	if imageID == "" {
+		fmt.Println("Can not find imageID")
+	}
+
+	return imageID, nil
```
DockToFuture: error return value is not needed please remove

PR #113, opened 2019-08-15
Minor nits

ialidzhikov:
- Use `New<Name>Cmd` funcs

Fixes #

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/cmd/info.go
+++ b/pkg/cmd/info.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"sort"
 
 	clientset "github.com/gardener/gardener/pkg/client/garden/clientset/versioned"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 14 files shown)


pkg/cmd/info.go
```diff
@@ -25,41 +25,42 @@ import (
...
+			Client, err = clientToTarget("garden")
+			gardenClientset, err := clientset.NewForConfig(NewConfigFromBytes(*kubeconfig))
+			checkError(err)
+			shootList, err := gardenClientset.GardenV1beta1().Shoots("").List(metav1.ListOptions{})
+			checkError(err)
+			fmt.Printf("Garden: %s\n", t.Target[0].Name)
+			fmt.Printf("Shoots:\n")
+			fmt.Printf("    total: %d \n", len(shootList.Items))
+			seedList, err := gardenClientset.GardenV1beta1().Seeds().List(metav1.ListOptions{})
+			for _, seed := range seedList.Items {
```
KristianZH: what do you think about the following algorithm. It will work faster:

```
shootsPerSeed := make(map[string]int)
	for _, shoot := range shootList.Items {
		shootsPerSeed[*shoot.Spec.Cloud.Seed]++
	}
	for seed, numberOfShoots := range shootsPerSeed {
		fmt.Printf("    %s: %d \n", seed, numberOfShoots)
	}
```

ialidzhikov: Nice suggestion. Applied it will little modification for sorting - to have always the seed seeds in output sorted (in the same order).

PR #114, opened 2019-08-22
Improve target command

KristianZH:

Currently the `gardenctl target seed/shoot` commands navigates to `/cache/seeds/`. This may cause a problem which I will describe bellow:

### Reproducing

1) Let's open a terminal and target some garden cluster and after that some seed cluster , for example `aws`.
  This will create a file in .garden/cache/seeds/aws/kubeconfig.yaml and will print it.
2) export KUBECONFIG=.garden/cache/seeds/aws/kubeconfig.yaml
3) Open another terminal where we will target again seed called `aws` but from another garden cluster. This will override the file, which points the Kubeconfig from the first terminal.

### Solution
We can just add the garden cluster's name to the Kubeconfig's path

Fixes #

@vpnachev @DockToFuture @ialidzhikov 

**Release note**:
-->
```improvement operator
None
```

### DIFF
--- a/pkg/cmd/download.go
+++ b/pkg/cmd/download.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/gardener/gardener/pkg/apis/garden/v1beta1"
@@ -59,6 +60,9 @@ func downloadTerraformFiles(option string) string {
 	var target Target
 	ReadTarget(pathTarget, &target)
 	Client, err = clientToTarget("garden")
+	gardenName := target.Stack()[0].Name
+	pathSeedCache := filepath.Join("cache", gardenName, "seeds")
+	pathProjectCache := filepath.Join("cache", gardenName, "projects")
 	if len(target.Target) < 3 && (option == "infra" || option == "internal-dns" || option == "external-dns" || option == "ingress" || option == "backup") {
 		fmt.Println("No Shoot targeted")
 		os.Exit(2)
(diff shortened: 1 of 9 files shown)


pkg/cmd/download.go
```diff
@@ -59,6 +59,9 @@ func downloadTerraformFiles(option string) string {
 	var target Target
 	ReadTarget(pathTarget, &target)
 	Client, err = clientToTarget("garden")
+	gardenName := target.Stack()[0].Name
+	pathSeedCache := "/cache/" + gardenName + "/seeds/"
```
ialidzhikov: You can use `filepath.Join(elem ...string)` everywhere.

pkg/cmd/download.go
```diff
@@ -134,6 +137,8 @@ func downloadLogs(option string) {
 	var target Target
 	ReadTarget(pathTarget, &target)
 	Client, err = clientToTarget("garden")
+	gardenName := target.Stack()[0].Name
```
ialidzhikov: You can move these 2 lines after the err check.

pkg/cmd/get.go
```diff
@@ -212,6 +212,7 @@ func getShoot(name string) {
 		}
 	}
 	Client, err = clientToTarget("garden")
+	gardenName := target.Stack()[0].Name
```
ialidzhikov: You can move this right after the target read operation, do not put it between another err check.

pkg/cmd/logs.go
```diff
@@ -444,12 +444,13 @@ func logsMachineControllerManager() {
 func logsKubernetesDashboard() {
 	var target Target
 	ReadTarget(pathTarget, &target)
+	gardenName := target.Stack()[0].Name
```
ialidzhikov: You can declare right before use. You can move to line 452.

pkg/cmd/operate.go
```diff
@@ -51,6 +51,7 @@ func operate(provider, arguments string) {
 	var target Target
 	ReadTarget(pathTarget, &target)
 	Client, err = clientToTarget("garden")
+	gardenName := target.Stack()[0].Name
```
ialidzhikov: You can move to line 72.

KristianZH:
Thank you @ialidzhikov for the comments. Can you check if the changes are matching your suggestions.

KristianZH:
@DockToFuture Yes i agree with this, but everywhere means, I have to go through all of the commands and replace everything. This will be out of the content of this PR and I think that doing this like another PR will be better approach.

DockToFuture:
Sounds reasonable.

PR #115, opened 2019-08-26
Replace hard coded strings with filepath.Join

KristianZH:

Replace hard coded paths with filepath.Join(). 
In this way, the paths will be relevant to Windows systems too.

Fixes #

@ialidzhikov @DockToFuture 

**Release note**:

```improvement operator
NONE
```


pkg/cmd/target.go
```diff
@@ -543,6 +507,43 @@ func targetShoot(targetWriter TargetWriter, shoot v1beta1.Shoot) {
 
 	KUBECONFIG = shootKubeconfigPath
 	fmt.Println("KUBECONFIG=" + KUBECONFIG)
+
```
DockToFuture: Moving this code to the end of the function causes the target cmd to fail, please undo this change.

PR #119, opened 2019-09-02
Fix panic in info cmd when Shoot is unscheduled

ialidzhikov:
Fix panic in info cmd occurring when Shoot `.spec.cloud.<provider>.seed` is `nil`. These entries are now shown as `unscheduled` in the output.

Fixes #

**Release note**:

```improvement operator
NONE
```

### DIFF
--- a/pkg/cmd/info.go
+++ b/pkg/cmd/info.go
@@ -15,57 +15,63 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
-	"io/ioutil"
-	"os"
 	"sort"
 
-	clientset "github.com/gardener/gardener/pkg/client/garden/clientset/versioned"
 	"github.com/spf13/cobra"
-	yaml "gopkg.in/yaml.v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/cmd/info.go
```diff
@@ -15,57 +15,63 @@
...
+
 			var sortedSeeds []string
 			for seed := range shootsCountPerSeed {
 				sortedSeeds = append(sortedSeeds, seed)
 			}
 			sort.Strings(sortedSeeds)
 
+			fmt.Fprintf(ioStreams.Out, "Garden: %s\n", targetStack[0].Name)
+			fmt.Fprintf(ioStreams.Out, "Shoots:\n")
+			fmt.Fprintf(ioStreams.Out, "    total: %d\n", len(shootList.Items))
```
vpnachev:
```suggestion
			fmt.Fprintf(ioStreams.Out, "\ttotal: %d\n", len(shootList.Items))
```

Oops, I did not publish the comment.
Anyway, you could use tabs `\t` instead of spaces.

ialidzhikov: Applied.

PR #138, opened 2019-11-08
improve `ssh` command to work with not joined node

KristianZH:
This PR fixes 2 issues:
 - 1: Ssh for Azure did not work because  `Start Instance` requires `zone-id` if we want 
       automation. The same `zone-id` is required for `aliyun ecs StopInstance`
 - 2: Now will be possible to SSH on node, which does not join the cluster.

Fixes #

@vpnachev @ialidzhikov @DockToFuture 

**Release note**:

```improvement operator
The `ssh` command now works with node name
```


pkg/cmd/ssh_gcp.go
```diff
@@ -131,7 +137,7 @@ func (g *GCPInstanceAttribute) createBastionHostInstance() {
 	_, err = tmpfile.Write(g.UserData)
 	checkError(err)
 
-	arguments := fmt.Sprintf("gcloud compute instances create %s --network %s --subnet %s --metadata-from-file startup-script=%s", g.BastionHostName, g.VpcName, g.Subnetwork, tmpfile.Name())
+	arguments := fmt.Sprintf("gcloud compute instances create %s --network %s --subnet %s --zone %s --metadata-from-file startup-script=%s", g.BastionHostName, g.VpcName, g.Subnetwork, g.Zone, tmpfile.Name())
```
DockToFuture: Please also add the zone flag in line 150 ```arguments = fmt.Sprintf("gcloud compute instances describe %s --zone %s --flatten=[status]", g.BastionHostName, g.Zone)``` and line 158```arguments := fmt.Sprintf("gcloud compute instances describe %s --zone %s --flatten=networkInterfaces[0].accessConfigs[0].natIP", g.BastionHostName, g.Zone) ``` as stated here.

DockToFuture:
Would be also nice if we could reflect the ip address behind the node name in the ```gardenctl ssh``` cmd. In aws it is refelected in the node name but for aliyun, azure and gcp we don't see it in the output. Would be nice to add a separated output column to the cmd.

vpnachev:
> Would be also nice if we could reflect the ip address behind the node name in the `gardenctl ssh` cmd. In aws it is refelected in the node name but for aliyun, azure and gcp we don't see it in the output. Would be nice to add a separated output column to the cmd.

@DockToFuture this is possible only for machines that have the respective `Node` objects in the cluster.
Basically, I do not see why the internal IP is required. The flow should be as following:
1. `gardenerctl ssh` will work only for shoot clusters and it will take the name of the `Machine` object as an argument. It must not depend on the respective `Node` object because it may not exist.
1. `gardenctl shell` will work with any k8s cluster that has `Node` objects - e.g. seed, shoot, garden. It will use the names of the `Node` objects. Here the internal IP can be printed as some additional info, but I would not support the IP as an argument of the command.

DockToFuture:
@vpnachev  We never talked about the ip as an argument, we just want to print it to ease debugging.

vpnachev:
Sure, but let it be only for `gardenctl shell`, not for `gardenctl ssh`.

pkg/cmd/ssh.go
```diff
@@ -82,16 +85,9 @@ func NewSSHCmd(reader TargetReader, ioStreams IOStreams) *cobra.Command {
 			}
 
 			if len(args) == 0 {
-				fmt.Printf("Node ips:\n")
-				printNodeIPs(kind)
+				fmt.Printf("Nodes\n")
```
DockToFuture: Please add a colon after Nodes.

DockToFuture: Rest looks good!

KristianZH:
> Sure, but let it be only for `gardenctl shell`, not for `gardenctl ssh`.

We discussed this suggestion with @DockToFuture and decided to not implement it for now, because there is no good approach for implementation.

PR #158, opened 2020-01-22
Add sessions to gardenctl

DockToFuture:
Sessions are added to `gardenctl`. This allows to operate on multiple clusters at the same time in different terminal sessions which was not possible before. The session ID can be set via the `GARDEN_SESSION_ID` environment variable.

**Release note**:

```improvement operator
Sessions are added to `gardenctl`. This allows to operate on multiple clusters at the same time in different terminal sessions which was not possible before. The session ID can be set via the `GARDEN_SESSION_ID` environment variable.
```


petersutter:
Can you elaborate what the purpose of the "session" is? This is not clear for me when reading the description / release note

Issue #160, opened 2020-02-03
Inform in case newer version is available


petersutter:
It would be nice if gardenctl would inform me if a new version is available, similar like other tools do it. And maybe also give instructions on how to update.

neo-liang-sap:
the latest release version could be obtained via (i took shell for example)
```
curl --silent "https://api.github.com/repos/gardener/gardenctl/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/'
```
returns `v0.16.0`

i will investigate how to implement sth similar in go for `gardenctl`.

please assign this ticket to me, thanks!

PR #167, opened 2020-03-09
issue 160 - add upgrade notification when new version available

neo-liang-sap:

This PR enables notification to end user when a new version of `gardenctl` is available while running `gardenctl version` command
Fixes #

https://github.com/gardener/gardenctl/issues/160

1) i added "new version notification" to command `gardenctl version` as the value of `VERSION` is injected in `gardenctl version` command only
2) to compare version numbers (in gardenctl it's e.g. `v0.15.0` or `v0.16.0`) i use one library `github.com/mcuadros/go-version`

**Release note**:

```improvement user
If a newer version of gardenctl is available, command "gardenctl version" will show notification including where and how to upgrade.
```

### DIFF
--- a/pkg/cmd/version.go
+++ b/pkg/cmd/version.go
@@ -15,9 +15,13 @@
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
+	"net/http"
 	"runtime"
 
+	version_tool "github.com/mcuadros/go-version"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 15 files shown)


pkg/cmd/version.go
```diff
@@ -38,6 +42,38 @@ func NewVersionCmd() *cobra.Command {
...
+
+	data := make(map[string]interface{})
+	err1 := json.Unmarshal([]byte(body), &data)
+	if err1 != nil {
+		print(err1)
+	}
+	if data["tag_name"] != nil {
+		latestVersion = data["tag_name"].(string)
+	}
+	if version_tool.CompareSimple(latestVersion, version) == 1 {
```
ialidzhikov: Why we use another package for version comparison when there is `github.com/Masterminds/semver` and it is used in g/g?

Issue #165, opened 2020-03-02
enable `gardenctl` to support select namespace while targeting garden/seed/shoot


neo-liang-sap:
Currently with 0.16 version of `gardenctl` we can target sth like `gardenctl target -g <GARDEN> -s <SEED> -t <SHOOT>`
It would be nice to add an switch of `-n namespace` to select the namespace

neo-liang-sap:
Oh - looks like i can't assign this issue to myself ;) could someone assign this to me? or add me permission in this repo to take issues ;)
my username is `neo-liang-sap`
Thanks!

petersutter:
done

neo-liang-sap:
Hi @petersutter :

I looked into `gardenctl target` related code and have some thoughts on this issue, do correct me if i was wrong in understanding this issue.

Currently there are 4 kinds of `target` - garden/seed/project/shoot - see https://github.com/gardener/gardenctl/blob/master/pkg/cmd/types.go#L73 , `namespace` is not an kind of `target`

Talking about `namespace` personally i think this should be kind of native concept in `kubectl`, it is meaningful to switch `namespace` while performing `kubectl` action, just like we switch `garden target` while performing `gardenctl` action

I see there's already an `kn` implementation for `kubectl` to switch namespace in https://github.com/gardener/gardenctl/blob/master/pkg/cmd/kubectl.go#L76 - which equals `kubectl --namespace=xxx`

* Do we still need to implement `gardenctl target -n namespace` as `gardenctl kn` already implemented this?

* If needed, what will be the desired outcome result if we implement `gardenctl target -n namespace`? 
Now in my debugging of shoot i use
1) `gardenctl target -g xxx -s xxx`
2) `kubectl` to switch namespace to `shoot_techinical_id`
So the desired implementation is 
`gardenctl target -g xxx -s xxx -n namespace` = 1) + 2)?

Do correct me if i was in wrong understanding of these above...Thanks!

cc @dansible

petersutter:
> * Do we still need to implement `gardenctl target -n namespace` as `gardenctl kn` already implemented this?

The problem with this is that you can't combine it into one command, so I guess your second point is desired outcome, if possible.

So from my point of view, when running
`gardenctl target -g xxx -s xxx -n namespace`
and then
`gardenctl k get pods` it would list the pods of the targeted namespace

neo-liang-sap:
PR - https://github.com/gardener/gardenctl/pull/168

PR #168, opened 2020-03-10
add support for gardenctl target namespace xxx or gardenctl target -m xxx

neo-liang-sap:

This PR enables `gardenctl target` command to switch to one namespace after target garden/seed/project/shoot

`gardenctl target namespace kube-system`
`gardenctl target -n kube-system`

and then you can get pods in `kube-system` in `gardenctl k get pods`

Fixes #
https://github.com/gardener/gardenctl/issues/165

* Personally i don't think my implementation of this issue is nice, as 
** namespace is kind of concept in k8s native, meanwhile `target` is native concept for gardener, this command just for make life easier compared with you need to explicitly write the name of namespace each time like `gardenctl k get pods -- -n shoot_techinical_id`
** in `gardenctl target` it deals target in api approach but i didn't find an solution to handle namespace in same api approach so i commandline to switch namespace, feel free give me some hint if this would be achieved in api way. 

**Release note**:

```improvement operator
Enable "gardenctl target" command to target namespace like "gardenctl target namespace xxx" or "gardenctl target -n xxx"
```

### DIFF
--- a/pkg/cmd/target.go
+++ b/pkg/cmd/target.go
@@ -35,20 +35,21 @@ import (
 const ProjectName = "project.garden.sapcloud.io/name"
 
 var (
-	pgarden  string
-	pproject string
-	pseed    string
-	pshoot   string
+	pgarden    string
+	pproject   string
+	pseed      string
+	pshoot     string
+	pnamespace string
 )
 
 // NewTargetCmd returns a new target command.
 func NewTargetCmd(targetReader TargetReader, targetWriter TargetWriter, configReader ConfigReader, ioStreams IOStreams) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:          "target <project|garden|seed|shoot> NAME",
+		Use:          "target <project|garden|seed|shoot|namespace> NAME",
 		Short:        "Set scope for next operations",
 		SilenceUsage: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if pgarden != "" || pproject != "" || pseed != "" || pshoot != "" {
+			if pgarden != "" || pproject != "" || pseed != "" || pshoot != "" || pnamespace != "" {
 				var arguments []string
 				if pgarden != "" {
 					arguments := append(arguments, "garden")
@@ -74,10 +75,15 @@ func NewTargetCmd(targetReader TargetReader, targetWriter TargetWriter, configRe
 					err := shootWrapper(targetReader, targetWriter, configReader, ioStreams, arguments)
 					checkError(err)
 				}
+
+				if pnamespace != "" {
+					err := setKubectlNamespace(pnamespace)
+					checkError(err)
+				}
 				return nil
 			}
-			if len(args) < 1 && pgarden == "" && pproject == "" && pseed == "" && pshoot == "" || len(args) > 4 {
-				return errors.New("command must be in the format: target <project|garden|seed|shoot> NAME")
+			if len(args) < 1 && pgarden == "" && pproject == "" && pseed == "" && pshoot == "" && pnamespace == "" || len(args) > 5 {
+				return errors.New("command must be in the format: target <project|garden|seed|shoot|namespace> NAME")
 			}
 			switch args[0] {
 			case "garden":
@@ -97,6 +103,14 @@ func NewTargetCmd(targetReader TargetReader, targetWriter TargetWriter, configRe
 				}
 			case "shoot":
 				err := shootWrapper(targetReader, targetWriter, configReader, ioStreams, args)
+				if err != nil {
+					return err
+				}
+			case "namespace":
+				if len(args) != 2 || args[1] == "" {
+					return errors.New("command must be in the format: target namespace NAME")
+				}
+				err := setKubectlNamespace(args[1])
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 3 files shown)


pkg/cmd/target.go
```diff
@@ -741,3 +760,26 @@ func shootWrapper(targetReader TargetReader, targetWriter TargetWriter, configRe
...
 	return nil
 }
+
+//set namespace for current kubectl ctx
+func setKubectlNamespace(kubectlNameSpace string) error {
+	if kubectlNameSpace == "" {
+		return errors.New("Namespace must be provided")
+	}
+
+	out1, err1 := exec.Command("bash", "-c", "gardenctl kubectl config current-context").Output()
```
DockToFuture: Use `ExecCmdReturnOutput`  with the the kubeconfig instead of calling the gardenctl binary.

pkg/cmd/target.go
```diff
@@ -741,3 +760,26 @@ func shootWrapper(targetReader TargetReader, targetWriter TargetWriter, configRe
...
+	if err1 != nil {
+		fmt.Println(err1)
+	}
+	currentConext := strings.TrimSuffix(string(out1), "\n")
+	fmt.Printf("Set namespace to %s for current context %s \n", kubectlNameSpace, currentConext)
+	cmdString := "gardenctl kubectl config set-context " + currentConext + " -- --namespace=" + kubectlNameSpace
```
DockToFuture: see above

pkg/cmd/target.go
```diff
@@ -196,16 +207,24 @@ func NewTargetCmd(targetReader TargetReader, targetWriter TargetWriter, configRe
...
 		},
-		ValidArgs: []string{"project", "garden", "seed", "shoot"},
+		ValidArgs: []string{"project", "garden", "seed", "shoot", "namespace"},
 	}
 
 	cmd.PersistentFlags().StringVarP(&pgarden, "garden", "g", "", "garden name")
 	cmd.PersistentFlags().StringVarP(&pproject, "project", "p", "", "project name")
 	cmd.PersistentFlags().StringVarP(&pseed, "seed", "s", "", "seed name")
 	cmd.PersistentFlags().StringVarP(&pshoot, "shoot", "t", "", "shoot name")
+	cmd.PersistentFlags().StringVarP(&pnamespace, "namespace", "m", "", "namespace name")
```
DockToFuture: Use `n` for namespace and replace `no-cache` option with `c`.

DockToFuture:
We could also add namespace to the target.Stack and treat it similar to garden, project, seed and shoot.
I would expect to keep the namespace until unset or until we change to a new garden, project, seed or shoot. Wdyt?

neo-liang-sap:
> We could also add the namespace to the target.Stack and treat it similar to garden, project, seed and shoot.
> I would expect gardenctl to keep the namespace until unset or until we change to a new target (garden, project, seed or shoot). Wdyt?

Totally agree with you @DockToFuture , i will refactor my code and implement all things including comments in above.

Thanks!

neo-liang-sap:
Hi @DockToFuture ,

I've modified following points based on your comments:
* https://github.com/gardener/gardenctl/pull/168#pullrequestreview-371724209 -> this is due to i didn't consider when `current-context` is empty, i've fixed this and each time setting context i will export kubeconfig based on current config
* use ExecCmdReturnOutput+kubeconfig to execute kubeconfig commands
* set -c for no-cache and -n for namespace

One thing i didn't fix is adding namespace to types of `target.Stack`. Previously i was trying to set this in `types.go` and use `TargetReader` to read current target and then `TargetWriter` to write current target.
Changing to this will have many influence on current all target related operations e.g. get target/ls target/drop target etc.
Besides that, all types of target.Stack refer to kind of `cluster`, garden/seed/shoot (project is not so handled differently), `namespace` is kind of concept within `cluster`.
My main concern is adding an relative small function may break current design and lots of other functions, what do you think?

Thanks
-Neo

pkg/cmd/target.go
```diff
@@ -100,6 +106,11 @@ func NewTargetCmd(targetReader TargetReader, targetWriter TargetWriter, configRe
 				if err != nil {
 					return err
 				}
+			case "namespace":
+				err := setKubectlNamespace(args[1])
```
DockToFuture: Also validate args[1] here, in case you enter `g target namespace` you will otherwise get a panic.

pkg/cmd/target.go
```diff
@@ -741,3 +760,29 @@ func shootWrapper(targetReader TargetReader, targetWriter TargetWriter, configRe
...
+//set namespace for current kubectl ctx
+func setKubectlNamespace(kubectlNameSpace string) error {
+
+	if kubectlNameSpace == "" {
+		return errors.New("Namespace must be provided")
+	}
+
+	currentConfig := getKubeConfigOfCurrentTarget()
+
+	out1, err1 := ExecCmdReturnOutput("bash", "-c", "export KUBECONFIG="+currentConfig+"; kubectl config current-context")
```
DockToFuture: you could use `out` and `err` instead off `out1` and `err1` and `out2` and `err2`

neo-liang-sap:
Hi @DockToFuture:
Thanks for the comments and i have modified according to your comments
-Neo

neo-liang-sap:
Hi @DockToFuture 
Thanks!
BTW i will create another issue to work on trying to PoC refactor namespace to be one kind of target

Thanks.
-Neo

PR #179, opened 2020-03-31
Move check for newer version to dedicated command

vpnachev:
Move check for newer version to dedicated command

It is very annoying `gardenctl version` to depend on network connectivity, thus moving the this check to dedicated command.

Fixes #

**Release note**:

```noteworthy user
The check for newer version is now moved to dedicated sub-command, e.g. `gardenctl update-check`.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/ghodss/yaml v1.0.0
 	github.com/golang/mock v1.3.1
 	github.com/jmoiron/jsonq v0.0.0-20150511023944-e874b168d07e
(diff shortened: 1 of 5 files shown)


DockToFuture:
Thank you for the PR, makes absolutely sense.
Would be great if we could also show an output if the latest version of gardenctl is installed like "gardenctl is up to date". What do you think?

vpnachev:
>  Would be great if we could also show an output if the latest version of gardenctl is installed like "gardenctl is up to date". What do you think?

Done.

Issue #164, opened 2020-02-26
Target garden by apiServerUrl


petersutter:
It would be nice if I could target the `garden` by providing the `apiServerUrl` / `server` as alternative to the name.

Background: In the dashboard on the cluster's details page an operator should be able to copy the `gardenctl` commands to
- target the shoot
  - Example:`gardenctl target --server https://example.com --seed the-seed --shoot my-shoot`) 
- target the seed of this shoot
  - Example: `gardenctl target --server https://example.com --seed the-seed`)

As the name of the garden varies for each local setup this could be way to uniquely identify the garden.

This could save some time typing and targeting the correct garden, shoot / seed.

In order for `gardenctl` to find the correct garden, it should check the `server` field in the kubeconfigs referenced in the respective gardens that matches the value of the `server` parameter. 
If either the kubeconfig or the `server` parameter is an ip and the other one a hostname, it first has to do a nslookup before comparison, of course.

petersutter:
with https://github.com/gardener/gardenctl/issues/165 one could also target the namespace in addition, like `gardenctl target --server https://example.com --seed the-seed --namespace shoot-technical-id`

neo-liang-sap:
some internal discussion with @petersutter ;)

PR #180, opened 2020-04-01
add support for gardenctl target server / gardenctl target -r

neo-liang-sap:
add support for gardenctl target server / gardenctl target -r
the server should be in format "https://servername:443" or "https://ipaddress:443"
Fixes https://github.com/gardener/gardenctl/issues/164
`gardenctl target server https://hostname:443`
`gardenctl target -r https://ipaddress:443 -s xxx -n xxx`
**Release note**:

```improvement operator
Added support for `gardenctl target server <uri>` / `gardenctl target -r <uri>` / `gardenctl target --server <uri>` to target a garden without knowing the garden name which can be different on each local installation
- The uri may be an ip or hostname
```

### DIFF
--- a/pkg/cmd/target.go
+++ b/pkg/cmd/target.go
@@ -18,6 +18,8 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -29,6 +31,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
 // ProjectName is they key of a label on namespaces whose value holds the project name.
@@ -40,17 +43,29 @@ var (
 	pseed      string
 	pshoot     string
 	pnamespace string
+	pserver    string
 )
 
 // NewTargetCmd returns a new target command.
-func NewTargetCmd(targetReader TargetReader, targetWriter TargetWriter, configReader ConfigReader, ioStreams IOStreams) *cobra.Command {
+func NewTargetCmd(targetReader TargetReader, targetWriter TargetWriter, configReader ConfigReader, ioStreams IOStreams, kubeconfigReader KubeconfigReader) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:          "target <project|garden|seed|shoot|namespace> NAME",
+		Use:          "target <project|garden|seed|shoot|namespace|server> NAME",
 		Short:        "Set scope for next operations",
 		SilenceUsage: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if pgarden != "" || pproject != "" || pseed != "" || pshoot != "" || pnamespace != "" {
+			if pgarden != "" || pproject != "" || pseed != "" || pshoot != "" || pnamespace != "" || pserver != "" {
 				var arguments []string
+				if pgarden != "" && pserver != "" {
+					fmt.Println("server and garden values can't be specified at same time.")
+					os.Exit(2)
+				}
+				if pserver != "" {
+					if !isValidURI(pserver) {
+						fmt.Println("the server must be a valid uri")
+						os.Exit(2)
+					}
+					serverWrapper(configReader, pserver, kubeconfigReader)
+				}
 				if pgarden != "" {
 					arguments := append(arguments, "garden")
 					arguments = append(arguments, pgarden)
@@ -82,8 +97,8 @@ func NewTargetCmd(targetReader TargetReader, targetWriter TargetWriter, configRe
 				}
 				return nil
 			}
-			if len(args) < 1 && pgarden == "" && pproject == "" && pseed == "" && pshoot == "" && pnamespace == "" || len(args) > 5 {
-				return errors.New("command must be in the format: target <project|garden|seed|shoot|namespace> NAME")
+			if len(args) < 1 && pgarden == "" && pproject == "" && pseed == "" && pshoot == "" && pnamespace == "" && pserver == "" || len(args) > 5 {
+				return errors.New("command must be in the format: target <project|garden|seed|shoot|namespace|server> NAME")
 			}
 			switch args[0] {
 			case "garden":
@@ -114,6 +129,21 @@ func NewTargetCmd(targetReader TargetReader, targetWriter TargetWriter, configRe
 				if err != nil {
 					return err
 				}
+			case "server":
+				if len(args) != 2 || args[1] == "" {
+					return errors.New("command must be in the format: target server NAME")
+				}
+				if !isValidURI(args[1]) {
+					fmt.Println("the server name must be a valid uri")
+					os.Exit(2)
+				}
+				serverWrapper(configReader, args[1], kubeconfigReader)
+				argStr := []string{"garden", pgarden}
+				err := gardenWrapper(targetReader, targetWriter, configReader, ioStreams, argStr)
+				if err != nil {
+					return err
+				}
+
 			default:
 				target := targetReader.ReadTarget(pathTarget)
 				if len(target.Stack()) < 1 {
@@ -220,7 +250,7 @@ func NewTargetCmd(targetReader TargetReader, targetWriter TargetWriter, configRe
 			}
 			return nil
 		},
-		ValidArgs: []string{"project", "garden", "seed", "shoot", "namespace"},
+		ValidArgs: []string{"project", "garden", "seed", "shoot", "namespace", "server"},
 	}
 
 	cmd.PersistentFlags().StringVarP(&pgarden, "garden", "g", "", "garden name")
@@ -228,6 +258,7 @@ func NewTargetCmd(targetReader TargetReader, targetWriter TargetWriter, configRe
 	cmd.PersistentFlags().StringVarP(&pseed, "seed", "s", "", "seed name")
 	cmd.PersistentFlags().StringVarP(&pshoot, "shoot", "t", "", "shoot name")
 	cmd.PersistentFlags().StringVarP(&pnamespace, "namespace", "n", "", "namespace name")
+	cmd.PersistentFlags().StringVarP(&pserver, "server", "r", "", "server name")
 
 	return cmd
 }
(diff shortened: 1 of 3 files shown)


pkg/cmd/target.go
```diff
@@ -739,6 +769,70 @@ func gardenWrapper(targetReader TargetReader, targetWriter TargetWriter, configR
...
+	}
+
+	if pgarden == "" {
+		fmt.Println("the server address enter could not be found as valid Garden target")
+		os.Exit(2)
+	}
+
+}
+
+func findIP(input string) string {
```
petersutter, line 802: instead, use https://golang.org/pkg/net/http/cookiejar/?m=all#isIP

pkg/cmd/target.go
```diff
@@ -739,6 +769,70 @@ func gardenWrapper(targetReader TargetReader, targetWriter TargetWriter, configR
...
+func checkServerName(input string) bool {
+	if strings.HasPrefix(input, "https://") && strings.HasSuffix(input, ":443") {
+		return true
+	}
+	return false
+}
+
+func getServerValueFromKubeconfig(kubeconfigPath string) string {
+	var kc Kubeconfig
+	yamlFile, err := ioutil.ReadFile(kubeconfigPath)
```
petersutter, line 819: use this code to read the kubeconfig https://github.com/gardener/gardenctl/blob/master/pkg/cmd/get.go#L163-L166
it probably makes sense to also move it to some utility function

pkg/cmd/target.go
```diff
@@ -739,6 +769,70 @@ func gardenWrapper(targetReader TargetReader, targetWriter TargetWriter, configR
...
+	checkError(err)
+	err = yaml.Unmarshal(yamlFile, &kc)
+	checkError(err)
+	if kc.Clusters[0].Cluster.Server == "" {
+		fmt.Println("can't get server value from kubeconfig")
+		os.Exit(2)
```
petersutter, line 825: not sure if the program should exit in this case or if it should just skip / ignore it in this case

pkg/cmd/target.go
```diff
@@ -739,6 +769,70 @@ func gardenWrapper(targetReader TargetReader, targetWriter TargetWriter, configR
...
+			kc := garden.KubeConfig
+			svr := getServerValueFromKubeconfig(kc)
+			if svr == serverName {
+				pgarden = garden.Name
+			}
+		}
+	}
+
+	if pgarden == "" {
+		fmt.Println("the server address enter could not be found as valid Garden target")
```
petersutter, line 796:
```suggestion
		fmt.Println("a garden could not be matched to the provided server address:", serverName)
```

pkg/cmd/target.go
```diff
@@ -739,6 +769,70 @@ func gardenWrapper(targetReader TargetReader, targetWriter TargetWriter, configR
...
+func findIP(input string) string {
+	numBlock := "(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])"
+	regexPattern := numBlock + "\\." + numBlock + "\\." + numBlock + "\\." + numBlock
+
+	regEx := regexp.MustCompile(regexPattern)
+	return regEx.FindString(input)
+}
+
+func checkServerName(input string) bool {
+	if strings.HasPrefix(input, "https://") && strings.HasSuffix(input, ":443") {
```
petersutter, line 811: I would rather only check if the uri can be parsed
```
import "net/url"

...

u, err := url.ParseRequestURI(input)
if err != nil {
...
}
```

pkg/cmd/target.go
```diff
@@ -40,17 +42,29 @@ var (
...
-			if pgarden != "" || pproject != "" || pseed != "" || pshoot != "" || pnamespace != "" {
+			if pgarden != "" || pproject != "" || pseed != "" || pshoot != "" || pnamespace != "" || pserver != "" {
 				var arguments []string
+				if pgarden != "" && pserver != "" {
+					fmt.Println("server and garden values can't be specified at same time.")
+					os.Exit(2)
+				}
+				if pserver != "" {
+					if !checkServerName(pserver) {
+						fmt.Println("the server name must be in https://hostname:443 or https://ipaddress:443")
```
petersutter, line 63: see comment below, should be a valid uri

pkg/cmd/target.go
neo-liang-sap: thanks, i've modified per your suggestion and moved this function to utils.go

pkg/cmd/target.go
neo-liang-sap: thanks, i removed 'os.Exit(2)'

pkg/cmd/target.go
```diff
@@ -739,6 +769,70 @@ func gardenWrapper(targetReader TargetReader, targetWriter TargetWriter, configR
...
 }
 
+func serverWrapper(reader ConfigReader, serverName string) {
+	config := reader.ReadConfig(pathGardenConfig)
+
+	if findIP(serverName) != "" {
+		for _, garden := range config.GardenClusters {
+			kc := garden.KubeConfig
+			svr := getServerValueFromKubeconfig(kc)
+			addrIP, err := net.LookupIP(removeHTTPSandPort(svr))
```
petersutter, line 779: use url.ParseRequestURI to get the fields you are interested in, e.g.
```
u, err := url.ParseRequestURI(svr)
checkError(err)
addrIP, err := net.LookupIP(u.Host)
```

pkg/cmd/target.go
neo-liang-sap: thanks, i've modified per your suggestion

pkg/cmd/target.go
neo-liang-sap: a bit wired in using the function isIP()
i see it in net/http/cookie package so i did:
import "net/http/cookie"
...
isIP(string) ->didn't pass compile, maybe i used wrong package?

Anyway i checked isIP implementation, the behind logic is `net.ParseIP(str)!=nil`

I tried `net.ParseIP(str)!=nil` to use as if, but in net.ParseIP function, it only returns not nil when the `str` is pure IP address like `127.0.0.1`, it will return nil if the str is `https://127.0.0.1` or `https://127.0.0.1:443`

for my scenario the usage might be https://hostname:443 or https://ipaddress:443, that's why i wrote an function for this....

welcome for suggestion

pkg/cmd/target.go
neo-liang-sap: > https://github.com/gardener/gardenctl/pull/180#discussion_r401579780
> https://github.com/gardener/gardenctl/pull/180#discussion_r401580188
> https://github.com/gardener/gardenctl/pull/180#discussion_r401601726

Hi @petersutter ,

For these three comments i'd like to have some discussion on this.
The core problem is to compare whether `gardenctl target server serverString` serverString matches server value in robot.kubeconfig.

I assume the server value in robot.kubeconfig is https://hostname:443 format, right? correct me if i was wrong.

My initial design for serverString are both https://hostname:443 and https://ipaddress:443.

So you would like to have following acceptable serverString:

1) hostname
2) ipaddress
3)hostname:port -> the port must be 443, am i right?
4)ipaddress:port
5) https://hostname:443
6) https://ipaddress:443
7) https://hostname
8) https://ipaddress

am i right? 

if above 8 scenario are all possible input then i will try to refactor my code to accept these 8 kinds of inputs.

Thanks.
-Neo

pkg/cmd/utils.go
```diff
@@ -218,3 +219,19 @@ func FetchShootFromTarget(target TargetInterface) (*gardencorev1beta1.Shoot, err
 
 	return shoot, nil
 }
+
+//get server field value from kubeconfig
+func getServerValueFromKubeconfig(kubeconfigPath string, kubeconfigReader KubeconfigReader) string {
+	var kc Kubeconfig
+	if strings.Contains(kubeconfigPath, "~") {
+		kubeconfigPath = filepath.Clean(filepath.Join(HomeDir(), strings.Replace(kubeconfigPath, "~", "", 1)))
+	}
```
petersutter, line 228: I just meant to move the highlighted part to a utility function to not have that much duplicate code.
E.g. if you search for the string `"~", "", 1)))` within the project you get a lot of duplicate / similar code that could be extracted to some utility function
This is only a suggestion to make the code cleaner

pkg/cmd/target.go
```diff
@@ -739,6 +769,57 @@ func gardenWrapper(targetReader TargetReader, targetWriter TargetWriter, configR
 	return nil
 }
 
+func serverWrapper(reader ConfigReader, serverName string, kubeconfigReader KubeconfigReader) {
```
petersutter, line 772: I will make a suggestion for this function as it does not cover the cases I had in mind..

pkg/cmd/utils.go
neo-liang-sap: Ah ok, got your point.
I will move the function `getServerValueFromKubeconfig` back to `target.go`.

For refactor of these three lines of code to an function and clean whole project , i will do in another PR

pkg/cmd/target.go
```diff
@@ -739,6 +769,57 @@ func gardenWrapper(targetReader TargetReader, targetWriter TargetWriter, configR
...
+	regEx := regexp.MustCompile(regexPattern)
+	return regEx.FindString(input)
+}
+
+func checkServerName(input string) bool {
+	if strings.HasPrefix(input, "https://") && strings.HasSuffix(input, ":443") {
+		return true
+	}
+	return false
+}
```
petersutter, line 815:
```suggestion
func isValidUri(input string) bool {
	_, err := url.ParseRequestURI(input)
	return err == nil
}
```

pkg/cmd/target.go
petersutter: Hi @neo-liang-sap, see PR for your fork https://github.com/neo-liang-sap/gardenctl/pull/1

pkg/cmd/target.go
petersutter: I hope https://github.com/neo-liang-sap/gardenctl/pull/1  covers all these cases, PTAL

pkg/cmd/target.go
petersutter: suggestion included in https://github.com/neo-liang-sap/gardenctl/pull/1

pkg/cmd/target.go
petersutter: suggestion included in https://github.com/neo-liang-sap/gardenctl/pull/1

pkg/cmd/utils.go
petersutter: thanks

pkg/cmd/target.go
petersutter: this suggestion is already included in https://github.com/neo-liang-sap/gardenctl/pull/1

pkg/cmd/target.go
petersutter: sorry, you are right you can't use what I suggested initially. Meanwhile I updated my comment and the suggestion is included in https://github.com/neo-liang-sap/gardenctl/pull/1

neo-liang-sap:
Hi @petersutter ,
thanks a lot for your kind suggestion - i've merged your PR into my branch and i did following two modification:
1) move function `getServerValueFromKubeconfig` back to target.go -> i will follow up with new issue to refactor "three lines" to an function and refactor whole project
2) rename isValidUri to isValidURI -> this is per go suggestion in my vs code...

Please kindly review current status of this PR and do let me know if i need further modification on this.

If all OK i will squash into one commit before merge.

Thanks again for your kind help in https://github.com/neo-liang-sap/gardenctl/pull/1

-Neo

neo-liang-sap:
thanks @petersutter , i've squash commits into one.

pkg/cmd/target.go
```diff
@@ -739,6 +769,137 @@ func gardenWrapper(targetReader TargetReader, targetWriter TargetWriter, configR
...
+func getServerValueFromKubeconfig(kubeconfigPath string, kubeconfigReader KubeconfigReader) string {
+	var kc Kubeconfig
+	if strings.Contains(kubeconfigPath, "~") {
+		kubeconfigPath = filepath.Clean(filepath.Join(HomeDir(), strings.Replace(kubeconfigPath, "~", "", 1)))
+	}
+	kubeconfig, err := kubeconfigReader.ReadKubeconfig(kubeconfigPath)
+	checkError(err)
+	err = yaml.Unmarshal(kubeconfig, &kc)
+	checkError(err)
+	if kc.Clusters[0].Cluster.Server == "" {
```
petersutter, line 897: I had another look at the code. Please take the server of the CurrentContext

neo-liang-sap: do you mean current-context in following screenshot? `kc.current-context`
this current-context is some kind of serviceaccount, not server name.

image

petersutter: on second thought, you could create a ClientConfig like here
https://github.com/gardener/gardenctl/blob/master/pkg/cmd/miscellaneous.go#L108-L115
and get the host from the config 
`Host string`
https://godoc.org/k8s.io/client-go/rest#Config

petersutter: > do you mean current-context in following screenshot? `kc.CurrentContext `

yes, the current-context points to a context name. As said, you do not have to bother with it, just use the ClientConfig (see initialization here https://github.com/gardener/gardenctl/blob/master/pkg/cmd/miscellaneous.go#L108-L115) to read the Host.

neo-liang-sap: thanks, i've modified per your suggestion, also squash into one commit.

pkg/cmd/target.go
```diff
@@ -739,6 +770,140 @@ func gardenWrapper(targetReader TargetReader, targetWriter TargetWriter, configR
 	return nil
 }
 
+func serverWrapper(reader ConfigReader, serverName string, kubeconfigReader KubeconfigReader) {
+	config := reader.ReadConfig(pathGardenConfig)
+
+	for _, garden := range config.GardenClusters {
+		kc := garden.KubeConfig
```
DockToFuture, line 777: Add a check to see if file path exists if not skip and go on with next
```go
if _, err := os.Stat(kc); os.IsNotExist(err) {
			break
		}
```

neo-liang-sap: thanks @DockToFuture , i've fixed this

petersutter: break or continue?

DockToFuture: `continue` i mean, thanks @petersutter  for checking

neo-liang-sap: thanks both, i've fixed and squash into one commit

DockToFuture:
Thanks a lot for the PR @neo-liang-sap .

Issue #188, opened 2020-04-20
gardenctl target target -u


rfranzke:
Can we support targeting a cluster by providing the Gardener Dashboard URL?
For example:

```bash
gardenctl target -u https://<url-to-dashboard>/namespace/garden-project/shoots/my-shoot/
```

neo-liang-sap:
sure i will look into it, please assign this issue to me , thanks!

rfranzke:
Thanks Neo!

neo-liang-sap:
PR submitted in https://github.com/gardener/gardenctl/pull/190

PR #190, opened 2020-04-20
add support for target url or target -u with dashboard url

neo-liang-sap:
Enable `gardenctl target dashboardUrl dashboard_url` or `gardenctl target -u dashboard_url` for directly targeting garden and shoot

This requires your (virtual) garden config to have `dashboardUrl` field set.
E.g.
```yaml
gardenClusters:
  - kubeConfig: /path/to/kubeconfig.yaml
    name: foo
    dashboardUrl: https://dashboard.foo.example.com
```

Fixes https://github.com/gardener/gardenctl/issues/188

the url could be in following two formats:
- https://<url-to-dashboard>/namespace/garden-project/shoots/my-shoot/
- https://<url-to-dashboard>/namespace/garden-project/shoots/my-shoot

**Release note**:
```improvement operator
Enable `gardenctl target dashboardUrl <dashboard_url>` or `gardenctl target -u <dashboard_url>` for directly targeting garden and shoot. Make sure to have `dashboardUrl` field set for each of your (virtual) garden configurations
```

### DIFF
--- a/pkg/cmd/target.go
+++ b/pkg/cmd/target.go
@@ -22,6 +22,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
@@ -38,22 +39,23 @@ import (
 const ProjectName = "project.garden.sapcloud.io/name"
 
 var (
-	pgarden    string
-	pproject   string
-	pseed      string
-	pshoot     string
-	pnamespace string
-	pserver    string
+	pgarden       string
+	pproject      string
+	pseed         string
+	pshoot        string
+	pnamespace    string
+	pserver       string
+	pdashboardurl string
 )
 
 // NewTargetCmd returns a new target command.
 func NewTargetCmd(targetReader TargetReader, targetWriter TargetWriter, configReader ConfigReader, ioStreams IOStreams, kubeconfigReader KubeconfigReader) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:          "target <project|garden|seed|shoot|namespace|server> NAME",
+		Use:          "target <project|garden|seed|shoot|namespace|server|dashboardUrl> NAME",
 		Short:        "Set scope for next operations",
 		SilenceUsage: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if pgarden != "" || pproject != "" || pseed != "" || pshoot != "" || pnamespace != "" || pserver != "" {
+			if pgarden != "" || pproject != "" || pseed != "" || pshoot != "" || pnamespace != "" || pserver != "" || pdashboardurl != "" {
 				var arguments []string
 				if pgarden != "" && pserver != "" {
 					fmt.Println("server and garden values can't be specified at same time.")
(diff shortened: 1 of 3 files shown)


neo-liang-sap:
tested in following scenraio , sorry for my bad drawing skills
image

pkg/cmd/target.go
```diff
@@ -1020,6 +1040,49 @@ func namespaceWrapper(targetReader TargetReader, targetWriter TargetWriter, kube
...
+func urlWrapper(urlString string) error {
+	gardenString := ""
+	shootString := ""
+	u, err := url.Parse(urlString)
+	if err != nil {
+		fmt.Println("the URL enter is invalid")
+		os.Exit(2)
+	}
+
+	if strings.Contains(u.Host, "canary") {
```
rfranzke, line 1054: Don't hard-code landscape namespace here. What if my landscape is named "foo"?

neo-liang-sap: good question, is there any regular pattern between url string and garden name?

if the URL is https://dashboard.garden.foo.xxx.xxx.xxx/ then the garden name will be foo-virtual?

petersutter: no, I don't think that you can derive the garden name from the url so that it works for all cases
I would also not hardcode the landscape types like dev, staging etc. This is only valid for SAP's landscapes.

Maybe the best thing to do is to put the dashboard url in the config for each garden? This way you can uniquely identify the corresponding garden, without having any assumptions on the naming schemas. WDYT?
```yaml
gardenClusters:
  - kubeConfig: /path/kubeconfig1.yaml
    name: foo
    dashboardUrl: https://mydashboard.example.com
  - kubeConfig: /path/kubeconfig2.yaml
    name: bar
    dashboardUrl: https://bar.example.com
```

rfranzke: +1 @petersutter

neo-liang-sap: i will open some internal discussion thread as for protecting sensitive information by email

neo-liang-sap: now the gardenconfig looks like this
image
 and i've modify this PR to use dashboard URL to identify garden name
thanks!

neo-liang-sap:
Hi @petersutter ,
Thanks, i've fixed all issues according to your comments.
-Neo

pkg/cmd/target.go
```diff
@@ -44,16 +45,17 @@ var (
 	pshoot     string
 	pnamespace string
 	pserver    string
+	purl       string
```
petersutter, line 48: pdashboardUrl

neo-liang-sap: thanks, i've fixed this

pkg/cmd/target.go
```diff
@@ -345,6 +366,22 @@ func resolveNameGarden(reader ConfigReader, name string) (matches []string) {
...
+func resolveGardenNameFromURL(reader ConfigReader, dashboardURL string) string {
+	config := reader.ReadConfig(pathGardenConfig)
+	gardenName := ""
+	for _, garden := range config.GardenClusters {
+		if strings.Contains(garden.DashboardURL, dashboardURL) {
+			gardenName = garden.Name
+		}
+	}
+	if gardenName == "" {
+		fmt.Println("Provided URL could not match a garden")
```
petersutter, line 379:
```suggestion
		fmt.Println("a garden could not be matched for the provided dashboard url")
```
to be in line with one of the other error messages (`a garden could not be matched for the provided server address`)

neo-liang-sap: thanks, i've fixed this.

petersutter:
@neo-liang-sap I have also updated your description and release note. PTAL

neo-liang-sap:
thanks @petersutter for fixing the description!

Issue #192, opened 2020-04-21
Improve handling of external resources


vpnachev:
**What would you like to be added**:
From time to time `gardenctl` is leaking cloud provider resources which is hard to find out and delete later.
An example is `gardenctl ssh <node-name>` which is ensuring the required node is accessible via pubic internet by creating a public IP, and depending on the cloud provider creates a bastion machine. If the ssh session is not gracefully deleted - the cloud resources are leaked and the end users might not be able to purge them manually. 

To prevent resource leaking, a CRD can be used to track these resources and a controller that is taking care to manage them, including to delete them properly. Something very similar and already implemented is the [terminal-controller](https://github.com/gardener/terminal-controller-manager). 

**Why is this needed**:
Prevent cloud resource leaking.

tedteng:
I will take a look ,  you can assign it to me

tedteng:
after investigation there are the following scenarios will cause the leaking and other issues. 

* scenarios 1) status checks timeout https://github.com/gardener/gardenctl/issues/191
Event instance status is running, till need to wait the status check passed before ec2 instance able to access.  some of case status check may take more than 60s 

Error
```
Waiting 60 seconds until ports are open.
Warning: Permanently added '<some-public-ip>' (ECDSA) to the list of known hosts.
channel 0: open failed: connect failed: Connection timed out
stdio forwarding failed
kex_exchange_identification: Connection closed by remote host
exit status 255
```

* scenarios 2) status checks not passed or other issues maybe leads the issues too.  
* scenarios 3) User A , B  access cluster for troubleshooting.
User A access creates `bastion ec2 A`, SG-bastion which includes rules host(22) and bastion(22) after executed `gardenctl ssh xxxx`.  graceful exit deletes everything normally.  but User B online this time. 

User A exit normally, `bastion ec2 A` , SG-bastion rule host(22)/bastion(22) has been deleted, `bastion ec2 B`  exist caused the dependent object can't delete SG-bastion
Error
```
  (3/3) Clean up bastion host secursity group

  Waiting 45 seconds until instance is deleted to remove all dependencies.

An error occurred (DependencyViolation) when calling the DeleteSecurityGroup operation: resource sg-032dfba7cf9b03b7a has a dependent object
exit status 2
````

there are two options that lead to two issues. 
* option 1
[...]

Issue #191, opened 2020-04-21
gardenctl ssh does not work


rfranzke:
**Describe the bug**
`gardenctl ssh` does not work:

```
$ gardenctl ssh ip-10-242-23-212.eu-central-1.compute.internal

Warning:
Be aware that you are entering an untrusted environment!
Do not enter credentials or sensitive data within the ssh session that cluster owners should not have access to.

Downloaded id_rsa key

(1/4) Fetching data from target shoot cluster
Data fetched from target shoot cluster.

(2/4) Setting up bastion host security group
Bastion host security group set up.
Opened SSH Port on Node.

(3/4) Creating bastion host
Bastion host instance created.

Instance State: pending

Instance State: pending

Instance State: pending

Instance State: running

Waiting 60 seconds until ports are open.
Warning: Permanently added '<some-public-ip>' (ECDSA) to the list of known hosts.
channel 0: open failed: connect failed: Connection timed out
stdio forwarding failed
kex_exchange_identification: Connection closed by remote host
exit status 255
(4/4) Cleanup
Cleaning up bastion host configurations...

Starting cleanup

  (1/3) Cleaning up bastion host instance
TERMINATINGINSTANCES	i-123456
CURRENTSTATE	32	shutting-down
PREVIOUSSTATE	16	running

  (2/3) Close SSH Port on Node.
  Closed SSH Port on Node.

  (3/3) Clean up bastion host security group

  Waiting 45 seconds until instance is deleted to remove all dependencies.

Bastion host configurations successfully cleaned up.
```

**To Reproduce**
`gardenctl ssh <nodename>`

**Expected behavior**
It should work.

**Gardenctl Version (please complete the following information):**
```
$ gardenctl version
gardenctl:
		version     : v0.17.0
		build date  : 2020-04-02
		go version  : go1.13.7
		go compiler : gc
		platform    : darwin/amd64
```

tedteng:
Hi rafael. May I know which shoot you are access it

rfranzke:
It was an AWS shoot. I guess it doesn't matter which. Take any AWS shoot.

tedteng:
OK I see I will take a look this issue

tedteng:
em I am take one of aws shoot it looks good from my side 

···
(⎈ |shoot--garden--aws-eu6:default)➜  kc.wiki git:(master) gardenctl ssh ip-10-250-0-33.eu-central-1.compute.internal

Warning:
Be aware that you are entering an untrusted environment!
[...]

PR #195, opened 2020-04-27
gardenctl ssh improvement

tedteng:
* fix: check if bastion host exists, avoid multiple bastions create and occupied source in cloud 
* add: waiting `instance-status.status,Values=ok`  then ssh  (ssh timeout due to status check initiating)
image

```
Warning: Permanently added '<some-public-ip>' (ECDSA) to the list of known hosts.
channel 0: open failed: connect failed: Connection timed out
stdio forwarding failed
kex_exchange_identification: Connection closed by remote host
exit status 255
```

* add: use exist global flags -c trigger clear up function  `gardenctl ssh xxxx -c`
* add: check ssh rule if exist avoid blocked the deletion flow.
```
(2/3) Close SSH Port on Node.

An error occurred (InvalidPermission.NotFound) when calling the RevokeSecurityGroupIngress operation: The specified rule does not exist in this security group.
```

Fixes https://github.com/gardener/gardenctl/issues/191 
Fixes https://github.com/gardener/gardenctl/issues/192

global flags testing
`gardenctl ssh (any hostname) -c` 
image

`gardenctl ssh and exit`
image

**Release note**:

```improvement operator
improvement operator handling of external resources relate with bastion instance in was, manual gracefully deleted by global flags enable
```

### DIFF
--- a/pkg/cmd/ssh.go
+++ b/pkg/cmd/ssh.go
@@ -93,7 +93,7 @@ func NewSSHCmd(reader TargetReader, ioStreams IOStreams) *cobra.Command {
 			case "azure":
 				sshToAZNode(args[0], path, user, pathSSKeypair, sshPublicKey)
 			case "alicloud":
-				sshToAlicloudNode(args[0], path, user, sshPublicKey)
+				sshToAlicloudNode(args[0], path, user, pathSSKeypair, sshPublicKey)
 			case "openstack":
 			default:
 				return fmt.Errorf("infrastructure type %q not found", infraType)
--- a/pkg/cmd/ssh_alicloud.go
+++ b/pkg/cmd/ssh_alicloud.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"strings"
 	"time"
 
@@ -69,7 +68,7 @@ type AliyunInstanceTypeSpec struct {
 }
 
 // sshToAlicloudNode provides cmds to ssh to alicloud via a public ip and clean it up afterwards.
-func sshToAlicloudNode(nodeName, path, user string, sshPublicKey []byte) {
+func sshToAlicloudNode(nodeName, path, user, pathSSKeypair string, sshPublicKey []byte) {
 	// Check if this is a cleanup command
 	if nodeName == "cleanup" {
 		cleanupAliyunBastionHost()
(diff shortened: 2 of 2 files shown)


vpnachev:
This PR is not fixing #192.

tedteng:
> This PR is not fixing #192.

@vpnachev , the current fix is able to remove bastion resources include (ec2 , sg, 22 rules) manually if there any resource leaking.  in https://github.com/gardener/gardenctl/issues/192 do you mean I need to refactor `ssh`, call `CRD` create `term-` pod then get through `term-` pod to access node? Is that idea want to achieve it?

vpnachev:
My idea is `gardenctl` to use `machine.sapcloud.io/v1alpha1.Machine` resource for the bastion host and let the [MCM](https://github.com/gardener/machine-controller-manager/) do the real work on creating/deleting these VMs. This approach is applicable for any cloud provider, not only AWS.

The implementation is, of course cloud dependent because on some providers, e.g. Azure, a public IP is directly attached to the VM and a bastion is not used.

tedteng:
more discuss in https://github.com/gardener/gardenctl/issues/192

tedteng:
sign-off

- no extra resources generated on Cloud Provider when using SSH Pod.
- able to SSH AWS, AZURE, GCP, ALICLOUD, OPENSTACK now
- `gardenctl ssh xxxxx` use Pod as jump box as default way.   flag `--j` will use Bastion VM as a jump box `gardenctl ssh --j xxxx`

- global flag `-c` also able to use for cleanup AWS Bastion VM `gardenctl ssh -c xxx`
- bugs fix
---

Openstack SSH Tesing result
image

Azure SSH Testing result
image

GCP SSH Testing result
image

AWS SSH Testing result
image

pkg/cmd/ssh_aws.go
```diff
@@ -64,15 +64,21 @@ func sshToAWSNode(nodeName, path, user, pathSSKeypair string, sshPublicKey []byt
 	a.createBastionHostSecurityGroup()
 	fmt.Println("")
 
+	// if global flag -c is specified call cleanupAwsBastionHost
```
DockToFuture, line 67: I would rather name the flag `cleanup` to make things more clear. Maybe we can also document it in the [readme](https://github.com/gardener/gardenctl/blob/master/README.md). We only have this for `aws`. In my opinion it would also make sense to add this for the other infrastructures, what do you think?

pkg/cmd/ssh.go
```diff
@@ -84,25 +87,31 @@ func NewSSHCmd(reader TargetReader, ioStreams IOStreams) *cobra.Command {
...
+					return fmt.Errorf("infrastructure type %q not found", infraType)
+				}
+			} else {
+				sshToPod(args[0], path, user, pathSSKeypair, sshPublicKey)
+			}
 			return nil
 		},
 	}
 
+	cmd.Flags().BoolVar(&bastionVM, "j", false, "Use Bastion VM as jump box")
```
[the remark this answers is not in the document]
tedteng: I am using local flags that seem not to support it. 
```
too many arguments in call to cmd.Flags().BoolVar
	have (*bool, string, string, bool, string)
	want (*bool, string, bool, string)go
```

pkg/cmd/ssh_aws.go
tedteng: sure.
 due to -c already use by no-cache, I change to -e as erase

tedteng: I agree with you, regarding the other infrastructures. I will work on that.

PR #202, opened 2020-05-29
Add github action to update gardenctl homebrew-tap

zanetworker:
Add a github action to update gardenctl  homebrew-tap

**Note to reviewer** 
We should create a secret token able to do workflow updates

**Release note**:

```improvement operator
An update to homebrew-tap is now automatically triggered. 
```

### DIFF
--- a/.github/workflows/update-gardenctl.yml
+++ b/.github/workflows/update-gardenctl.yml
@@ -1,0 +1,17 @@
+name: gardenctl-updater
+
+on:
+  release:
+    types:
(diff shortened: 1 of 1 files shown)


neo-liang-sap:
following up with @zanetworker

zanetworker:
a secret named `ACCESS_TOKEN` needs to be created in this repo with the following permissions `read:packages, repo, workflow, write:packages` on other gardener repositories. cc @msohn

PR #219, opened 2020-07-09
Add terraform subcommand

### LINKED ISSUE #134: Add a terraform subcommand
**What would you like to be added**:
We should add scripts that can ease work with terraform states.

**Why is this needed**:
Sometimes there are errors caused by terraform job pods:
1. A resource has already been created by a previous execution but the terraform state was not properly saved
1. There are resource leaks due to multiple executions of terraform jobs

To fix these the operator has to either manually delete the created resources on the cloud provider or manually update the terraform state configmaps kept in the Shoot's namespace on the Seed.

To ease this work we can introduce a couple of scripts which take care of automatically setting up the terraform work directory and properly updating the state's configmap.

jfortin-sap:
Add terraform subcommand to debug cluster terraform issues
Fixes #134 

**Release note**:

```improvement operator
Add terraform subcommand to debug cluster terraform issues
```

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -139,7 +139,6 @@ func init() {
 		NewTargetCmd(targetReader, targetWriter, configReader, ioStreams, kubeconfigReader),
 		NewDropCmd(targetReader, targetWriter, ioStreams),
 		NewGetCmd(targetReader, configReader, kubeconfigReader, kubeconfigWriter, ioStreams))
-	RootCmd.AddCommand(NewTerraformCmd(targetReader))
 	RootCmd.AddCommand(NewDownloadCmd(), NewShowCmd(), NewLogsCmd())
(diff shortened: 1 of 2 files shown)


pkg/cmd/root.go
```diff
@@ -139,6 +139,7 @@ func init() {
 		NewTargetCmd(targetReader, targetWriter, configReader, ioStreams, kubeconfigReader),
 		NewDropCmd(targetReader, targetWriter, ioStreams),
 		NewGetCmd(targetReader, configReader, kubeconfigReader, kubeconfigWriter, ioStreams))
+	RootCmd.AddCommand(NewTerraformCmd(targetReader))
```
DockToFuture, line 142: Please add it after `RootCmd.AddCommand(NewKubectxCmd())`.

Issue #217, opened 2020-07-07
gardenctl can not support aws format kubeconfig


knightXun:
it's seems gardenctl will read some content from kubeconfig, and judge the format whether is illeage.I see some code of gardenctl and compare this with my aws format kubeconfig, finding gardenctl can not support aws kubeconfig very well. Please notice this.

neo-liang-sap:
gardenctl can't support some kind of kubeconfig e.g. containing use with command line content.
say if your kubeconfig user part contains `aws-iam-authenticator` command line to generate user info, this could not be consumed by gardenctl

btw i found some misunderstanding content in your description which i can't edit them, e.g. `weather `->`wether`, `finding gardenctl can support aws kubeconfig very well`->`finding gardenctl can't support aws kubeconfig very well`, correct me if i was wrong.

thanks

knightXun:
@neo-liang-sap thanks.

Issue #175, opened 2020-03-19
Not possible to use OIDC based kubeconfig files for authentication 


donistz:
**Describe the bug**
It is not possible to use gardenctl for operation of seed clusters where the K8S API Server is configured with OIDC based authentication. 
**Expected behavior**
It should be possible to configure the usage of OIDC based kubeconfig files for accessing the Gardener seed clusters.

vpnachev:
There is an issue in gardener about generating kubeconfig with OIDC context - https://github.com/gardener/gardener/issues/1433. It should ease support in gardenctl about OIDC.

neo-liang-sap:
Hi @DockToFuture / @vpnachev 
Currently in oidc enabled scenario, `gardenctl` is not able to target `shoot`, i'm not sure whether i'm using correctly as follows, will this issue https://github.com/gardener/gardenctl/issues/175 fix this? or https://github.com/gardener/gardener/issues/1433 will fix?
CC @dansible 
Thanks!

![image](https://user-images.githubusercontent.com/42594392/85349733-d3396380-b531-11ea-8c0b-d87e6e231d01.png)

neo-liang-sap:
@DockToFuture as https://github.com/gardener/gardenctl/pull/221 is closed and merged i believe this one could also be closed?
thanks
-Neo

DockToFuture:
Thanks for noticing. Closed with #221.

PR #221, opened 2020-07-14
remove exec and auth provider check to utilize kubeconfig with oidc enabled

neo-liang-sap:
In gardenctl there's validation against kubeconfig which doesn't allow kubeconfig contains `exec` or `auth provider` to prevent malicious executable code. Now with OIDC enabled, kubeconfig contains exec part, so i would like to propose this PR to discuss whether we can remove this check in gardenctl 

Fixes https://github.com/gardener/gardenctl/issues/217 and https://github.com/gardener/gardenctl/issues/175

/CC @dansible @ThormaehlenFred @DockToFuture @ialidzhikov @vpnachev @donistz 

Basically this PR enables gardenctl to utilize oidc enabled kubeconfig, i did some basic testing like `gardenctl target garden/seed/shoot` and then do some kubectl operation like `gardenctl k get pods -- -n kube-system` etc , so far so good.

Indeed it is security to allow using kubeconfig with exec part which can execute any code, as discussed in https://banzaicloud.com/blog/kubeconfig-security/ (thanks for @DockToFuture ), we could discuss e.g. we add some warning message in gardenctl project regarding this? to let people know gardenctl allow using kubeconfig contains exec part which could be security risk....

**Release note**:

```improvement operator
remove kubeconfig validation for exec and auth provider, to utilize kubeconfig with oidc enabled.
```

### DIFF
--- a/pkg/cmd/utils.go
+++ b/pkg/cmd/utils.go
@@ -164,7 +164,7 @@ func NewConfigFromBytes(kubeconfig string) *restclient.Config {
 // ValidateClientConfig validates that the auth info of a given kubeconfig doesn't have unsupported fields.
 func ValidateClientConfig(config clientcmdapi.Config) error {
 	validFields := []string{"client-certificate-data", "client-key-data", "token", "username", "password"}
-
+	pathOfKubeconfig := getKubeConfigOfCurrentTarget()
 	for user, authInfo := range config.AuthInfos {
 		switch {
 		case authInfo.ClientCertificate != "":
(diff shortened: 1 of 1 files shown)


petersutter:
I think it depends on whether the kubeconfigs are coming / are read from a trusted source.
- Is the garden landscape that you are targeting trusted?
- Have you got the kubeconfig for the garden landscape from a trusted source

If the above is true, we _could_ assume that also the kubconfig secrets that we read from the gardener are safe. The gardener has implemented a [validating webhook](https://github.com/gardener/gardener/blob/e852d53f1745c3221fe05487e076eb0a2bb164c8/pkg/controllermanager/server/handlers/webhooks/validate_kubeconfig_secrets.go) to [validate](https://github.com/gardener/gardener/blob/9248b6e42f251a1c22f0f20dfce32075246df6cf/pkg/client/kubernetes/client.go#L172-L193) secrets containing kubeconfigs

DockToFuture:
Yes, I also agree with @petersutter here. It's always good to double check but I guess we gain more out of it if we leave it to the user to check the `kubeconfig` if it was not downloaded from a trusted source. What does the rest of you think any objections?

neo-liang-sap:
thanks for your comments @petersutter @DockToFuture , i had another suggestion, if we could allow `exec` in `kubeconfig` utilized by `gardenctl`, maybe i can add some warning message when the validation hook found there's exec part in kubeconfig? say some warning message like `Your kubeconfig contains exec part which could execute malicious code, please make sure your kubeconfig comes from trusted source.....`

What do you think? if it's OK i will add these warning message to this PR

Thanks

petersutter:
In case we want to display a warning message (do not have a strong opinion here)
- it should be clear which kubeconfig is meant, e.g. 
  - local path of the downloaded kubeconfig
  - and/or by naming it like:
    - shoot kubeconfig for `my-project/my-shoot`
    - seed kubeconfig `my-seed`
    - garden kubeconfig `xyz`
- a warning of course only makes sense if the kubeconfig is not used (a kubectl command is run) before the user had the chance to acknowledge/read the warning

> `Your kubeconfig contains exec part which could execute malicious code[...]`

nit: change to [...] could *contain* malicious code[...]

Other suggestion:
`Kubeconfig for [...] under path [...] contains exec and or auth provider configurations that could contain malicious code. Please only continue if you have verified it to be uncritical`

neo-liang-sap:
i added warning msg and test result looks like this
image

neo-liang-sap:
i updated PR with warning message like this
image

Issue #222, opened 2020-07-15
fetch Public IP and restrict the source IP range.


tedteng:
**Describe the bug** 
Base on security scan
bastions host will create security groups with 0.0.0.0/0 IP range allow to access as jumpbox. 
will enhance SSH feature, fetch user Public IP first then create security groups base on the user Public IP instead of 0.0.0.0/0.

PR #223, opened 2020-07-15
restrict the source IP range

tedteng:
fetch user Public IP and restrict the source IP range 
create ssh rules in security group base on user public IP when use `gardenctl ssh`

Fixes https://github.com/gardener/gardenctl/issues/222

**Release note**:
restrict the source IP range when use `gardenctl ssh`


Kristian-ZH:
Nice PR,

Can you please squash the commits because I do not see much value to separate them

pkg/cmd/ssh_alicloud.go
```diff
@@ -198,9 +199,9 @@ func (a *AliyunInstanceAttribute) createBastionHostSecurityGroup() {
 			os.Exit(2)
 		}
 		fmt.Println("Configuring bastion host security group rules...")
-		_, err = ExecCmdReturnOutput("bash", "-c", "aliyun ecs AuthorizeSecurityGroup --Policy Accept --NicType intranet --Priority 1 --SourceCidrIp 0.0.0.0/0 --PortRange 22/22 --IpProtocol udp --SecurityGroupId="+a.BastionSecurityGroupID)
+		_, err = ExecCmdReturnOutput("bash", "-c", "aliyun ecs AuthorizeSecurityGroup --Policy Accept --NicType intranet --Priority 1 --SourceCidrIp %s --PortRange 22/22 --IpProtocol udp --SecurityGroupId="+a.BastionSecurityGroupID, a.MyPublicIP)
```
Kristian-ZH, line 202: I think it will be better to extract this command in function which accepts `protocol`. 
It is difficult to see now where is the difference between these two `bash` commands

tedteng: will remove udp it seem SSH protocol no needed

Issue #229, opened 2020-07-23
increase stability of the existing ssh command


tedteng:
**Describe the bug**
`gardenctl ssh` not stability when connecting AWS. most of time exit with an error 
```
kex_exchange_identification: Connection closed by remote host
exit status 255
```

**To Reproduce**
 `gardenctl ssh nodename` 

**Expected behavior**

**Screenshots**

**Gardenctl Version (please complete the following information):**
gardenctl:
		version     : v0.18.0
		build date  : 2020-07-13
		go version  : go1.14.3
		go compiler : gc
		platform    : darwin/amd64

**Additional context**

PR #231, opened 2020-07-23
SSH code optimization

tedteng:
increase stability of the existing ssh and code optimization to fix the ssh unstable connection issue.

add bastion ssh port open detect before ssh exected.
even ssh open already sometimes AWS bastion still not ready to connect. so add ssh retry 3 time to fix the case
add prompt msg as verbosity when ssh,  the error msg from ssh will return issues from bastion or attend to connect nodes

image

Fixes https://github.com/gardener/gardenctl/issues/229

@danielfoehrKn 

**Release note**:
increase the stability of the existing ssh and code optimization.

### DIFF
--- a/pkg/cmd/ssh_aws.go
+++ b/pkg/cmd/ssh_aws.go
@@ -72,13 +72,27 @@ func sshToAWSNode(nodeName, path, user, pathSSKeypair string, sshPublicKey []byt
 	fmt.Println("(3/4) Creating bastion host")
 	a.createBastionHostInstance()
 
+	fmt.Println("Waiting SSH 22 port open")
+	attemptCnt := 0
+	for attemptCnt < 60 {
+		ncCmd := fmt.Sprintf("nc -rz -w 5 %s 22", a.BastionIP)
+		cmd := exec.Command("bash", "-c", ncCmd)
+		output, _ := cmd.CombinedOutput()
+		fmt.Println(" ")
+		if strings.Contains(string(output), "succeeded") {
+			break
+		}
+		time.Sleep(time.Second * 5)
+		attemptCnt++
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/cmd/ssh_aws.go
```diff
@@ -72,13 +72,27 @@ func sshToAWSNode(nodeName, path, user, pathSSKeypair string, sshPublicKey []byt
 	fmt.Println("(3/4) Creating bastion host")
 	a.createBastionHostInstance()
 
+	fmt.Println("Waiting SSH 22 port open")
```
danielfoehrKn, line 75: Maybe  in the future it makes sense to have a logging framework with configurable verbosity levels. 
For instance to also see verbose output of the SSH command (good for understanding what went wrong).
Just to keep in mind.

danielfoehrKn:
Thanks @neo-liang-sap !

pkg/cmd/ssh_aws.go
tedteng: sure np, at the moments, after fixing the ssh process display looks like this when ssh failure"  we can get brief information `closed by 3.121.114.80 port 22`  to identify the issue from local to bastion instance or bastion instance to node instance.

```
SSH gardener@3.121.114.80 => gardener@ip-10-250-23-169.eu-central-1.compute.internalConnection closed by 3.121.114.80 port 22
kex_exchange_identification: Connection closed by remote host
exit status 255
```

After investigation this time, I got some new findings. I think the root cause is from AWS EC2,  As testing, SSH services may not fully running even EC2 instance status become running from pending/Initializing..  which causes SSH command may be hung at the first time until timeout exit.  so I detecting 22 port open and retry ssh command this time. 


-Ted

PR #270, opened 2020-08-24
[WIP]move get seed out of targetShoot method

### LINKED ISSUE #269: remove getSeed from targetShoot
**Describe the bug**
current `targetShoot` method mix with operation get seed information. 

I think `target shoot` maybe only focus on pure shoot method?  During the code development and code review @petersutter and @dansible aware of this phenomenon as well.  

so I am planning to move get seed out of targetShoot method eg start from L564. 

It may be some history reasons or convenience (get seed log after targetshoot ?) put seed operation in `targetShoot` Let me know anything I need to be aware of or other concerns. 

`seed, err := gardenClientset.CoreV1beta1().Seeds().Get(*shoot.Spec.SeedName, metav1.GetOptions{})`

https://github.com/gardener/gardenctl/blob/c95543feccd02472106914abb334a3b655bb1abb/pkg/cmd/target.go#L553-L655

**Expected behavior**

**Screenshots**

 - latest version

**Additional context**

tedteng:

Fixes https://github.com/gardener/gardenctl/issues/269

**Release note**:

```improvement operator
code clean up  targetShoot
```

### DIFF
--- a/pkg/cmd/target.go
+++ b/pkg/cmd/target.go
@@ -556,23 +556,23 @@ func targetShoot(targetWriter TargetWriter, shoot gardencorev1beta1.Shoot, reade
 
 	// Get and cache seed kubeconfig for future commands
 	gardenName := target.Stack()[0].Name
-	pathSeedCache := filepath.Join(pathGardenHome, "cache", gardenName, "seeds")
+	// pathSeedCache := filepath.Join(pathGardenHome, "cache", gardenName, "seeds")
 	pathProjectCache := filepath.Join(pathGardenHome, "cache", gardenName, "projects")
 
-	gardenClientset, err := target.GardenerClient()
-	checkError(err)
-	seed, err := gardenClientset.CoreV1beta1().Seeds().Get(*shoot.Spec.SeedName, metav1.GetOptions{})
-	checkError(err)
+	// gardenClientset, err := target.GardenerClient()
+	// checkError(err)
+	// seed, err := gardenClientset.CoreV1beta1().Seeds().Get(*shoot.Spec.SeedName, metav1.GetOptions{})
+	// checkError(err)
 	gardenClient, err := target.K8SClientToKind(TargetKindGarden)
 	checkError(err)
-	seedKubeconfigSecret, err := gardenClient.CoreV1().Secrets(seed.Spec.SecretRef.Namespace).Get(seed.Spec.SecretRef.Name, metav1.GetOptions{})
-	checkError(err)
-	var seedCacheDir = filepath.Join(pathSeedCache, *shoot.Spec.SeedName)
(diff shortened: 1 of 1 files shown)


pkg/cmd/target.go
```diff
@@ -632,9 +632,10 @@ func targetShoot(targetWriter TargetWriter, shoot gardencorev1beta1.Shoot, reade
 
 	// Cache shoot kubeconfig
 	var shootCacheDir string
-	if target.Target[1].Kind == "seed" {
-		shootCacheDir = filepath.Join(pathSeedCache, target.Target[1].Name, shoot.Name)
-	} else if target.Target[1].Kind == "project" {
+	// if target.Target[1].Kind == "seed" {
```
neo-liang-sap, line 635: in short, this will causing problem for scenario like `gardenctl target -s xxx -t xxx`
There are many combination scenarios in target Stack, the `target.Target[0]` is garden for sure, otherwise code will  error, the `target.Target[1]` could be `seed`, `project`, `shoot` ,`namespace`.  I would suggest you perform full test (especially action related to seed ) against your PR

tedteng: I noticed there are a lot of combination scenarios when using target stack. also mention in another ticket full test required as well, We need to decide the user behavior at the beginning.  

Take the example, the case you mention "-s xxx -t xxx"  I am not understanding here what kind of user case needed and the purpose?  do we have a real case you need target seed then target shoot cluster again?  but definitely there will be more real cases we don't know or need be aware of. 


for my understanding and the design for now

to target seed:  gardenctl target -g xxx --seed soil-gcp --namespace shoot--garden--az-eu2
to target shoot: gardenctl target -g xxx --project garden --shoot az-eu2

neo-liang-sap: - "-s xxx -t xxx" is real case, say i want to target a shoot within a seed after target a garden
- there are many other scenarios maybe you are not aware of, e.g
 target shoot
   - gardenctl target shoot xxx (currently stack might be garden/seed , garden/project , garden )
   - gardenctl target xxx (without xxx type specified)

tedteng: still not got the point "-s xxx -t xxx".  to target seed or target shoot are different, that I already provider the command as above which you can also get from the access bar from the gardener console page.  


for the second part about the usage of target shoot whatever you are using target `shoot` or `-t` it is the same goal to target the shoot cluster. Even you mention currently stack might be garden/seed, garden/project, garden, for this part should not take care in `targetShoot` method. that why I suggest remove get seed or the thing you mention garden/seed, garden/project method out of `targetShoot` method.

tedteng:
dependency on https://github.com/gardener/gardenctl/pull/263 merge

PR #274, opened 2020-08-25
refactor curl to use net/http

neo-liang-sap:
**What this PR does / why we need it**:
refactor curl to use net/http
**Which issue(s) this PR fixes**:
Fixes https://github.com/gardener/gardenctl/issues/273

**Special notes for your reviewer**:

**Release note**:
```improvement developer
refactor curl to use net/http
```

PR #276, opened 2020-08-26
Ipv4 check

### LINKED ISSUE #275: ipv4 check after get public IP
**Describe the bug**
in some rare case public ip return ipv6. 

 ```
(2/4) Setting up bastion host security group
aws ec2 authorize-security-group-ingress --group-id sg-0ccef1d8a628a6a91 --protocol tcp --port 22 --cidr 240e:e0:f090:34bb:bd78:6c06:e328:9b79/32

An error occurred (InvalidParameterValue) when calling the AuthorizeSecurityGroupIngress operation: CIDR block 240e:e0:f090:34bb:bd78:6c06:e328:9b79/32
```

**To Reproduce**
gardenctl ssh node

**Expected behavior**
ipv4 return

**Screenshots**

 - Version/Commit [latest version]

**Additional context**

### PR DESCRIPTION
**Pull request by tedteng**

Fixes https://github.com/gardener/gardenctl/issues/275

**Release note**:

```improvement operator
Support valid Public IP for IPv4 and IPv6 in aws bastion security group when `ssh`

```

### DIFF
--- a/pkg/cmd/miscellaneous.go
+++ b/pkg/cmd/miscellaneous.go
@@ -21,10 +21,10 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
-	"strconv"
 	"strings"
 
 	"github.com/gardener/gardener/pkg/apis/core"
@@ -327,22 +327,8 @@ func capture() func() (string, error) {
 	}
 }
 
-func isIP(word string) bool {
-	parts := strings.Split(word, ".")
-	if len(parts) < 4 {
-		return false
-	}
-	for _, x := range parts {
-		if i, err := strconv.Atoi(x); err == nil {
-			if i < 0 || i > 255 {
-				return false
-			}
-		} else {
-			return false
-		}
-
-	}
-	return true
+func isIPv4(host string) bool {
+	return net.ParseIP(host) != nil && net.ParseIP(host).To4() != nil
 }
 
 func getPublicIP() string {
@@ -356,5 +342,9 @@ func getPublicIP() string {
 	defer resp.Body.Close()
 	ip, err := ioutil.ReadAll(resp.Body)
 	checkError(err)
+	if !isIPv4(string(ip)) {
+		fmt.Println("Not valid ipv4 address")
+		os.Exit(1)
+	}
 	return string(ip)
 }
--- a/pkg/cmd/ssh_aws.go
+++ b/pkg/cmd/ssh_aws.go
@@ -17,6 +17,7 @@ package cmd
 import (
 	"fmt"
 	"io/ioutil"
+	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -56,7 +57,7 @@ type AwsInstanceAttribute struct {
 func sshToAWSNode(nodeName, path, user, pathSSKeypair string, sshPublicKey []byte, myPublicIP string) {
 	a := &AwsInstanceAttribute{}
 	a.SSHPublicKey = sshPublicKey
-	a.MyPublicIP = myPublicIP + "/32"
+	a.MyPublicIP = myPublicIP
 
 	fmt.Println("")
 
@@ -171,7 +172,7 @@ func (a *AwsInstanceAttribute) createBastionHostSecurityGroup() {
 		return
 	}
 
-	// create security group and ssh rule
+	// create security group for bastion host
 	arguments := fmt.Sprintf("aws ec2 create-security-group --group-name %s --description ssh-access --vpc-id %s", a.BastionSecurityGroupName, a.VpcID)
 	captured := capture()
 	operate("aws", arguments)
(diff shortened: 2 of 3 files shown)


neo-liang-sap:
please hold this PR until https://github.com/gardener/gardenctl/pull/274 , currently under review from @petersutter for urlencoding

tedteng:
> please hold this PR until #274 , currently under review from @petersutter for urlencoding

sure, definitely this PR merge will wait for #274

tedteng:
find one issue before rebase, will waiting for the fix https://github.com/gardener/gardenctl/pull/274
image

tedteng:
remove /lgtm

pkg/cmd/ssh_aws.go
```diff
@@ -180,10 +180,14 @@ func (a *AwsInstanceAttribute) createBastionHostSecurityGroup() {
 	a.BastionSecurityGroupID = strings.Trim((capturedOutput), "\n")
 	arguments = fmt.Sprintf("aws ec2 create-tags --resources %s  --tags Key=component,Value=gardenctl", a.BastionSecurityGroupID)
 	operate("aws", arguments)
-	arguments = fmt.Sprintf("aws ec2 authorize-security-group-ingress --group-id %s --protocol tcp --port 22 --cidr %s", a.BastionSecurityGroupID, a.MyPublicIP)
+
+	if isIP(a.MyPublicIP) {
+		arguments = fmt.Sprintf("aws ec2 authorize-security-group-ingress --group-id %s --protocol tcp --port 22 --cidr %s/32", a.BastionSecurityGroupID, a.MyPublicIP)
+	} else {
+		arguments = fmt.Sprintf("aws ec2 authorize-security-group-ingress --group-id %s --ip-permissions IpProtocol=tcp,FromPort=22,ToPort=22,Ipv6Ranges=[{CidrIpv6=%s/64}]", a.BastionSecurityGroupID, a.MyPublicIP)
+	}
```
petersutter, line 188: `isIP` is a custom implemented function to check for ipv4 addresses. Replace the custom implementation by using `net.ParseIP` and `return netIP != nil && netIP.To4() != nil` to check for ipv4
see also https://github.com/kubernetes/kubernetes/blob/dc025534afb0c3106b3aab166f88b971102135d1/vendor/k8s.io/utils/net/net.go#L116-L125 as example for the ipv6 check

In the end the code should look something like:
if isIPv4String(a.MyPublicIP) {
  ...
} else if IsIPv6String(a.MyPublicIP) {
  ...
} else {
  // error
}

pkg/cmd/ssh_aws.go
```diff
@@ -180,10 +181,17 @@ func (a *AwsInstanceAttribute) createBastionHostSecurityGroup() {
 	a.BastionSecurityGroupID = strings.Trim((capturedOutput), "\n")
 	arguments = fmt.Sprintf("aws ec2 create-tags --resources %s  --tags Key=component,Value=gardenctl", a.BastionSecurityGroupID)
 	operate("aws", arguments)
-	arguments = fmt.Sprintf("aws ec2 authorize-security-group-ingress --group-id %s --protocol tcp --port 22 --cidr %s", a.BastionSecurityGroupID, a.MyPublicIP)
+
+	ip := net.ParseIP(a.MyPublicIP)
```
petersutter, line 185: If the string is not a valid textual representation of an IP address, ParseIP returns nil. Hence you also need to check for nil and that's why I suggested initially to move that code into utils/miscellaneous and dump/replace the `isIP` function that is already there

tedteng: update isIP method which return bool to check whether IP is valid for ipv4 or ipv6. 
also change to logic in here, add exit after print the "IP not valid:" + a.MyPublicIP.

pkg/cmd/miscellaneous.go
```diff
@@ -327,22 +327,8 @@ func capture() func() (string, error) {
...
-			}
-		} else {
-			return false
-		}
-
-	}
-	return true
+func isIP(host string) bool {
+	return net.ParseIP(host) != nil
 }
```
petersutter, line 332: Even if the function was named isIP, the old implementation would only return true in case it was an ipv4 address.
Now it will return true if it's an ipv6 and ipv4 address..

tedteng: should I rename isIP to like isIPv4 something like that then check and return only ipv4 address ?

petersutter: yes I guess this would be better. It's only used in one place currently `createBastionHostInstance` (gcp). Not sure if this will work with ipv6 addresses..

petersutter (approved):
/lgtm, thanks
however I haven't tested it

pkg/cmd/miscellaneous.go
tedteng: fixed, so far GCP `createBastionHostInstance`  not support ipv6 yet I think, but I will check it later when I work on support GCP host in the `user` role later.

pkg/cmd/miscellaneous.go
```diff
@@ -356,5 +342,9 @@ func getPublicIP() string {
 	defer resp.Body.Close()
 	ip, err := ioutil.ReadAll(resp.Body)
 	checkError(err)
+	if !isIP(string(ip)) {
```
petersutter, line 345:
```suggestion
	if !isIPv4(string(ip)) {
```
shouldn't this be isIPv4?

tedteng: yes thanks.

neo-liang-sap (changes requested):
1) this commit should not exist in this PR
image
2) these three commits are dummy content, should not exist in this PR
image

neo-liang-sap (approved):
/lgtm
however i haven't tested, confirmed with @tedteng he's fully confidence with it

PR #296, opened 2020-09-02
fix logic covering both garden+seed+shoot and garden+project+shoot

### LINKED ISSUE #295: gardenctl `aws|gcp...` rely that project will be targeted
**Describe the bug**
`gardenctl` expects that project will be targeted in cmd/operate.go
https://github.com/gardener/gardenctl/blob/master/pkg/cmd/operate.go#L47 

In this way if `gardenctl aws ...` is called and the Path to the shoot is constructed by using its seed, the command will fail with
```
projects.core.gardener.cloud "<seed-name>" not found
```

**To Reproduce**
g target garden... -> g target seed... -> g target shoot ...

**Expected behavior**
It should also expect that the Target can be constructed using `seed` instead of `project`

neo-liang-sap:
current operate.go only covers scenario when target stack is garden+project+shoot, should also cover scenario garden+seed+shoot
Fixes https://github.com/gardener/gardenctl/issues/295

/CC @Kristian-ZH 
**Release note**:
```improvement operator
Fix operate.go missing coverage on scenario garden+seed+shoot
```

### DIFF
--- a/pkg/cmd/operate.go
+++ b/pkg/cmd/operate.go
@@ -43,11 +43,28 @@ func operate(provider, arguments string) {
 
 	gardenClientset, err := target.GardenerClient()
 	checkError(err)
-
-	project, err := gardenClientset.CoreV1beta1().Projects().Get(target.Stack()[1].Name, metav1.GetOptions{})
-	checkError(err)
-	shoot, err := gardenClientset.CoreV1beta1().Shoots(*project.Spec.Namespace).Get(target.Stack()[2].Name, metav1.GetOptions{})
-	checkError(err)
+	var shoot *gardencorev1beta1.Shoot
+	if target.Stack()[1].Kind == "project" {
+		project, err := gardenClientset.CoreV1beta1().Projects().Get(target.Stack()[1].Name, metav1.GetOptions{})
+		checkError(err)
+		shoot, err = gardenClientset.CoreV1beta1().Shoots(*project.Spec.Namespace).Get(target.Stack()[2].Name, metav1.GetOptions{})
+		checkError(err)
+	} else if target.Stack()[1].Kind == "seed" {
+		shootList, err := gardenClientset.CoreV1beta1().Shoots("").List(metav1.ListOptions{})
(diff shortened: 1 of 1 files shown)


neo-liang-sap:
the test covers following flow:
1) gardenctl target -g xxx -s xxx -t xxx (garden target is garden+seed+shoot), then `gardenctl aws ec2 xxx ` works fine
2) gardenctl target -g xxx -p xxx -t xxx (garden target is garden+project+shoot), then `gardenctl aws ec2 xxx ` works fine

@Kristian-ZH i hope i'm understanding correctly on your issue, do let me know if i missed anything, thanks!

Kristian-ZH:
Yup, this will resolve the issue.

Thanks!

Kristian-ZH (approved):
/lgtm

pkg/cmd/operate.go
```diff
@@ -43,11 +43,28 @@ func operate(provider, arguments string) {
...
+		project, err := gardenClientset.CoreV1beta1().Projects().Get(target.Stack()[1].Name, metav1.GetOptions{})
+		checkError(err)
+		shoot, err = gardenClientset.CoreV1beta1().Shoots(*project.Spec.Namespace).Get(target.Stack()[2].Name, metav1.GetOptions{})
+		checkError(err)
+	} else if target.Stack()[1].Kind == "seed" {
+		shootList, err := gardenClientset.CoreV1beta1().Shoots("").List(metav1.ListOptions{})
+		checkError(err)
+		var filteredShoots []gardencorev1beta1.Shoot
+		for _, s := range shootList.Items {
+			if s.Name == target.Stack()[2].Name {
```
petersutter, line 57: @neo-liang-sap  sorry for beeing late for the review, but here you are missing the check for the seed. Do not add to `filteredShoots` if the seed does not match the one on the target stack

neo-liang-sap: thanks! i added a new PR for this [here](https://github.com/gardener/gardenctl/pull/298)

Issue #307, opened 2020-09-08
[refactor] remove deprecated `ReadTarget(pathTarget, &target)` and use `target := targetReader.ReadTarget(pathTarget)`


neo-liang-sap:
there are lots of occurrence across the project while reading the target
```
var target Target
ReadTarget(pathTarget, &target)
```
however this method is deprecated
```
// ReadTarget file into Target
// DEPRECATED: Use `TargetReader` instead.
```
the recommend usage to read target is
1) pass a `TargetReader ` to any command
2) in the command, use following to read the target
```
target := targetReader.ReadTarget(pathTarget)
```

neo-liang-sap:
will use this issue to track progress across project refactor

neo-liang-sap:
some of the old method has been refactored, some didn't due to limitations (the new method needs a TargetReader injection and not suits for all occasion e.g. target.go)
/close

PR #308, opened 2020-09-08
refactor download/terraform and add test case for download

### LINKED ISSUE #306: add test case for `gardenctl download` command
add test case for `gardenctl download` command

neo-liang-sap:
This PR contains things from many aspect:
- [x]  part of https://github.com/gardener/gardenctl/issues/307, contains refactor for `download` / `terraform`
- [x] added test case for `download`, in oder to make `download.go` to be testable i also make some modification to `download.go`
- [x] when `getRole() == "user"` , we can't assume current stack is `garden/project/shoot`, we should add check and fail the process when current stack is `garden/seed/shoot` , see https://github.com/gardener/gardenctl/compare/master...neo-liang-sap:download-test?expand=1#diff-5ea91c3cc672b9d1b4df67072a980d4aR66 , @tedteng FYI
- [x] as download.go involves terraform related actions, i also modified terraform.go, @jfortin-sap FYI
- [x] ssh.go is modified as ssh also involves terraform related action

Fixes https://github.com/gardener/gardenctl/issues/306

**Release note**:
```improvement operator
```

### DIFF
--- a/pkg/cmd/download.go
+++ b/pkg/cmd/download.go
@@ -15,6 +15,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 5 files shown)


pkg/cmd/download.go
```diff
@@ -30,42 +31,45 @@ import (
...
-	ReadTarget(pathTarget, &target)
+	target := targetReader.ReadTarget(pathTarget)
 	// return path allow non operator download key file
 	if getRole() == "user" {
-		gardenName := target.Stack()[0].Name
-		projectName := target.Stack()[1].Name
 		if (len(target.Stack()) < 3) || (len(target.Stack()) == 3 && target.Stack()[2].Kind == "namespace") {
 			fmt.Println("No Shoot targeted")
 			os.Exit(2)
+		} else if target.Stack()[1].Kind == "seed" {
```
tedteng, line 66: first of all, we don't have any guidelines in README to encourage `user` role to target -shoot via -seed,
secondly, the `user` role can't see the seed information form cluster WebUI as default.  In this way, I think the `user` role may don't have any chance to target seed in any case. so I am assuming this code maybe never executing for `user` role

neo-liang-sap: 1) I'm not saying we have anything `encourage` user action to target shoot, in line 64 `projectName := target.Stack()[1].Name` i don't think it's proper as we don't check whether target.Stack()[1] is seed or project but just directly assume it's project
2) in gardenctl , the `target` is read from a file , anyone including user role could modify the file to make the garden target include garden/seed/shoot, the user role is only determined during run time api call, so this piece code of check is necessary

tedteng: Besides, if you really want to add msg I think it should be in `target.go` file for `targetseed` part. as it is the beginning of the process when `user` role operator any task.

but as mentioned at the above, not seed information,  I guess, this code maybe not executing for the `user` role

neo-liang-sap: > Besides, if you really want to add msg I think it should be in `target.go` file for `targetseed` part. as it is the beginning of the process when `user` role operator any task.

this is a good point for enhancement of `target.go` but out of this PR scope

> but as mentioned at the above, not seed information, I guess, this code maybe not executing for the `user` role

as i mentioned above, anyone have access to local file got a chance to set target like `garden/seed/shoot`

tedteng: To generate the local file is base on rules from `target.go` and the input from user. It should be handle in there at the begging for everything not just fix in the `download.go`. That why I say the code in there is not necessary.  why not we just block and display the warming information in `user` role when they target seed at the begging instead of fix in `download.go`

neo-liang-sap: refactor `target.go` will not prevent from anyone modifying local file directly

tedteng: why people modifying local file directly ? besides, the logic for all the gardenctl reload from `.garden/sessions/plantingSession/target` ,  everytime when `user` role `target -g garden -p project -t shoot` will overwrite it by target.go.

neo-liang-sap: why can't ???

tedteng: > why can't ???

My suggestion is if you need to check `seed` you can add in `target.go` at the beginning of the process.

neo-liang-sap: > > why can't ???
> 
> My suggestion is if you need to check `seed` for `user` role you can add in `target.go` at the beginning of the process.

i don't think we are on same page here.
To Summary:
- for enhancement of target.go i've already opened new issue https://github.com/gardener/gardenctl/issues/309
- what i'm talking about is that you can't prevent people from editing target file ->you never know whether it's `g/s/s` or `g/p/s` in target stack -> you need to check this, that's what my PR stands for.

petersutter:
@neo-liang-sap this release not is not very helpful, especially not interesting for operators. Either change it to developer or remove it completely, which I would suggest in this case.

neo-liang-sap:
thanks @petersutter , i've removed the release notes

PR #310, opened 2020-09-09
refactor show.go and add test case for gardenctl show

neo-liang-sap:
- [x] refactor show.go for ReadTarget method
- [x] add test case for `gardenctl show`
Fixes #

**Release note**:

```improvement operator
```


petersutter:
also here, @neo-liang-sap this release not is not very useful, especially not interesting for operators. Either change it to developer or remove it completely, which I would suggest in this case.

neo-liang-sap:
thanks @petersutter , i've removed the release notes

PR #314, opened 2020-09-15
sanitize_267_ls_refactor

### LINKED ISSUE #267: Sanitize user input when executing local commands
User input should be sanitized before local command execution, otherwise there is a possibility for command injection.

E.g. see this code
https://github.com/gardener/gardenctl/blob/3ff05d6244a4cc19d618a55037dfd0ac1b9b31b8/pkg/cmd/ssh_alicloud.go#L202

a.MyPublicIP is determined by calling an external service 
https://github.com/gardener/gardenctl/blob/3ff05d6244a4cc19d618a55037dfd0ac1b9b31b8/pkg/cmd/miscellaneous.go#L327-L331

If this service returns a command instead of an IP, I assume that this command would get executed as per code above.
There are a lot of other places where local commands are executed and most of them, if not all, do not sanitize the input. 

And taking another look at the above example, why is the aliyun binary not called directly with the other parameters as arguments? https://github.com/gardener/gardenctl/blob/3ff05d6244a4cc19d618a55037dfd0ac1b9b31b8/pkg/cmd/ssh_alicloud.go#L202

See also https://github.com/gardener/gardenctl/issues/266 to further reduce the need to execute local commands and thereby reducing the attack vector.

tedteng:
sanitize bash -c from ls.go
Fixes https://github.com/gardener/gardenctl/issues/267

**Release note**:

```improvement operator

```

### DIFF
--- a/pkg/cmd/ls.go
+++ b/pkg/cmd/ls.go
@@ -348,8 +348,7 @@ func getSeedsWithShootsForProject(ioStreams IOStreams) {
 //getNamespaces get all namespaces based on current kubeconfig
 func getNamespaces(ioStreams IOStreams) {
 	currentConfig := getKubeConfigOfCurrentTarget()
-
-	out, err := ExecCmdReturnOutput("bash", "-c", "export KUBECONFIG="+currentConfig+"; kubectl get ns")
+	out, err := ExecCmdReturnOutput("kubectl", "--kubeconfig="+currentConfig, "get", "ns")
 	if err != nil {
 		fmt.Println(err)
 	}


pkg/cmd/ls.go
```diff
@@ -93,17 +93,11 @@ func NewLsCmd(targetReader TargetReader, configReader ConfigReader, ioStreams IO
 // printProjectsWithShoots lists list of projects with shoots
 func printProjectsWithShoots(target TargetInterface, ioStreams IOStreams) error {
```
petersutter, line 94:
```go
func printProjectsWithShoots(target TargetInterface, ioStreams IOStreams) error
```
The signature indicates that an error is returned. This isn't the case anymore. 

Also I'm not sure if I like the fact that in case of error the process exits and the caller is not able to handle the error. Maybe the caller is not interested in the error in this case.. but I would not introduce this as default error handling pattern for gardenctl. Just something to think about..

petersutter (changes requested):

pkg/cmd/ls.go
tedteng: forced update by mistake. issue fixed

petersutter (approved):
/lgtm

PR #318, opened 2020-09-18
refactor getTechnicalID

### LINKED ISSUE #312: Gardenctl ssh fails when targeting a non unique shoot name
**Describe the bug**

Gardenctl is unable to ssh into a node if there are multiple shoots in one gardener instance with the same name.
When running `gardenctl ssh` on a shoot with a non unique name, gardenctl is unable to determine the correct context and tries to access the seed of the other shoot: `Error: stat /Users/xxx/.garden/cache/live/seeds/other-seed/kubeconfig.yaml: no such file or directory`

**To Reproduce**
Create 2 shoots with the same name in different projects
1. target one of the shoots.
2. `gardenctl ssh`
3. `Error: stat /Users/xxx/.garden/cache/live/seeds/other-seed/kubeconfig.yaml: no such file or directory`

**Expected behavior**
Gardenctl should be able ssh into a node even if there are multiple shoots with the same name

 - v0.21.0

**Additional context**
cc @DockToFuture please correct or add your observations to the description.

tedteng:

Fixes https://github.com/gardener/gardenctl/issues/312

**Release note**:

```improvement operator|developer
fix ssh target shoots with the same name in different projects. 
refactor getTechnicalID become to light and efficiently fetch
```

### DIFF
--- a/pkg/cmd/miscellaneous.go
+++ b/pkg/cmd/miscellaneous.go
@@ -350,17 +350,23 @@ func getRole() string {
 }
 
 /*
-getTechnicalID returns the Technical Id, which used as clustername of the shoot cluster
+getTechnicalID returns the Technical Id for shoot or seed, used for namespace, clustername,
 */
-func getTechnicalID() string {
+func getTechnicalID(shootName string) string {
 	var target Target
 	ReadTarget(pathTarget, &target)
+	var out string
+	var err error
+	switch shootName {
+	case "shoot":
+		KUBECONFIG := getKubeConfigOfClusterType("shoot")
+		out, err = ExecCmdReturnOutput("kubectl", "--kubeconfig="+KUBECONFIG, "config", "view", "-o", "jsonpath={.clusters[0].name}")
+		checkError(err)
+	case "seed":
+		KUBECONFIG := getKubeConfigOfClusterType("seed")
+		out, err = ExecCmdReturnOutput("kubectl", "--kubeconfig="+KUBECONFIG, "config", "view", "-o", "jsonpath={.clusters[0].name}")
+		checkError(err)
+	}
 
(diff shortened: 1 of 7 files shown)


tedteng:
@petersutter  let me know what you think, I change to a new way to fetch the getTechnicalID base from kubeconfig, it more light and efficiently. let me know any concerns. if everything all good. I am planning replace funcation later. https://github.com/gardener/gardenctl/blob/e4e5fa2184075e9bd6c449d06f05ee5de924e89b/pkg/cmd/miscellaneous.go#L162

tedteng:
will check test case later

pkg/cmd/miscellaneous.go
```diff
@@ -350,17 +350,23 @@ func getRole() string {
...
+	}
 
-	gardenClientset, err := target.GardenerClient()
-	checkError(err)
-	project, err := gardenClientset.CoreV1beta1().Projects().Get(target.Stack()[1].Name, metav1.GetOptions{})
-	checkError(err)
-	shoot, err := gardenClientset.CoreV1beta1().Shoots(*project.Spec.Namespace).Get(target.Stack()[2].Name, metav1.GetOptions{})
-	checkError(err)
-	return shoot.Status.TechnicalID
+	return out
```
petersutter, line 371: I think this is a fragile solution, even though it is faster than fetching the technical id from the shoot resource. Do not rely on the cluster name in the kubeconfig. This is just a coincidence that it matches from my point of view.
I would go with a solution like already started in this PR https://github.com/gardener/gardenctl/pull/317

petersutter (changes requested):
I'm not in favour of taking the kubeconfig cluster name (with index 0) as technicalId. So a veto from my side. I would prefer the solution from https://github.com/gardener/gardenctl/pull/317

PR #344, opened 2020-09-24
fix ssh oidc issue

tedteng:
**What this PR does / why we need it**:
fix ssh oidc wariming msg block ssh issue

due to the oidc warming msg, the capture() will return more information cause the issue. for a quick fix. I comments the print. let me know if any issue then will take time wokring on later

``` 
"Kubeconfig under path /Users/i333878/.garden/cache/canary/projects/i333878/aws-users-role/kubeconfig.yaml contains exec configurations that could contain malicious code. Please only continue if you have verified it to be uncritical
sg-09527df1718b9a7b0
"
```

**Which issue(s) this PR fixes**:
Fixes https://github.com/gardener/gardenctl/issues/322

**Special notes for your reviewer**:

**Release note**:

```improvement operator
fix ssh oidc wariming msg block ssh issue
```

PR #353, opened 2020-09-28
rollback getTechnical in SSH

### LINKED ISSUE #322: gardenctl ssh returns `Error parsing parameter '--filters': Expected: '=', received: 'EOF' for input:` for AWS shoot cluster
**Describe the bug**
say we have a sample AWS cluster
use gardenctl ssh to one of the node
image

**To Reproduce**
1. Which target was set 'gardenctl get target'
```
 gardenctl get target                                                  
target:
- kind: garden
  name: xxxxxdev
- kind: project
  name: sretestneo
- kind: shoot
  name: neo-test-ar
```
2. Which command was entered [e.g. 'gardenctl show vpn-seed']
`gardenctl ssh ip-10-250-27-0.eu-west-1.compute.internal`
3. What was the output of the command 
```
Error parsing parameter '--filters': Expected: '=', received: 'EOF' for input:
under
     ^
```
**Expected behavior**
gardenctl ssh should success

**Screenshots**
see above
 - Version/Commit [e.g. 8cc4e0c3]
```
gardenctl version                                                     (ccee-m3/default)
gardenctl:
		version     : v0.22.0
		build date  : 2020-09-21
		go version  : go1.14.6
		go compiler : gc
		platform    : darwin/amd64
```

**Additional context**

i'm not sure but i suspect this is due to recent changes in `ssh_aws.go`
image

tedteng:
rollback change getTechnial in SSH

Fixes https://github.com/gardener/gardenctl/issues/322

**Release note**:

```improvement operator

```

### DIFF
--- a/pkg/cmd/miscellaneous.go
+++ b/pkg/cmd/miscellaneous.go
@@ -15,6 +15,7 @@
 package cmd
 
 import (
+	"bufio"
 	"errors"
 	"flag"
 	"fmt"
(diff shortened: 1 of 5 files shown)


petersutter (changes requested):
This is not an acceptable solution. It does not fix the root cause and in addition it re-introduces old code that was replaced with the new getTechnicalID implementation by a reason
- fetch all shoots across all namespaces (which is not necessary)
- fragile `if strings.HasSuffix(namespace.Name, clustername) {` check

tedteng:
> This is not an acceptable solution. It does not fix the root cause and in addition it re-introduces old code that was replaced with the new getTechnicalID implementation by a reason
> 
> * fetch all shoots across all namespaces (which is not necessary)
> * fragile `if strings.HasSuffix(namespace.Name, clustername) {` check

This PR only rollback change getTechnicalID to original. I believe this will change it again when `getFromTargetInfo` done.

To fix the root cause It will be in the ticket issue https://github.com/gardener/gardenctl/issues/354

neo-liang-sap (changes requested):
I agree with @petersutter , this is not acceptable.
@tedteng , please take your time to work on https://github.com/gardener/gardenctl/issues/322 and fix it from root cause, as i mentioned in https://github.com/gardener/gardenctl/pull/344#pullrequestreview-495276148 , we don't have tight release schedule here.
BTW i don't have any urgent testing blocked here, so again, take your time.

PR #401, opened 2020-10-20
test for kubectl get ns

### LINKED ISSUE #381: printout namespaces with api call instead of kubectl call
Replace this method https://github.com/gardener/gardenctl/blob/master/pkg/cmd/ls.go#L349-L356 with API call.

Related to https://github.com/gardener/gardenctl/issues/266

andrei-panov:
add test for *kubectl get ns* command
Fixes #381 

**Release note**:

```improvement operator

```


andrei-panov:
We should discuss if we want to detach Integration tests execution with a special flag or exclude them by flag.
And it's not clear if there any default kubeconfig provided during test execution.

PR #402, opened 2020-10-20
WIP: Deploy gardenctl Gardener shoot

jguipi:
 It adds a GitHub action pipeline that deploy gardener shoot manifest

### DIFF
--- a/.github/actions/k8s/action.yml
+++ b/.github/actions/k8s/action.yml
@@ -1,0 +1,20 @@
+name: 'Apply a k8s manifest'
+description: 'This actions will apply k8s manifest'
+inputs:
+  path-to-file:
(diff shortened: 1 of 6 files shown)


petersutter:
@jguipi out of curiosity, what is the purpose of those 4 clusters running on an internal landscape?

dansible:
@petersutter we need these clusters to be able to validate that `gardenctl ssh` works for each cloud provider. They are in our internal landscape as they are only short-lived and should hibernated as soon as the tests have completed

petersutter:
Ok got it.
One of the clusters has a different hibernation schedule than the others. Maybe you want to check?
Also, you maybe want to set the purpose to `testing`? In this case no monitoring stack is deployed.

PR #425, opened 2020-11-03
Issue 326 sanitize logs go

### LINKED ISSUE #326: Sanitize user input in logs.go
**Describe the bug**

Sanitize user input when executing local commands as much as possible

This is one of the sub-task for ticket https://github.com/gardener/gardenctl/issues/267

andrei-panov:
replace bash call with kubectl

Fixes #326 

It should be possible to follow commits to make the review. I left some methods and tests which will be deleted after review.
At the moment I can't test if it's really getting the logs because I'm missing some permission. 

**Release note**:
<!--  Write your release note:
NONE


pkg/subcmd/kubectl/kubectl.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	if sinceSeconds != 0 {
+		command.WriteString(fmt.Sprintf("--since=%vs ", sinceSeconds.Seconds()))
+	}
+
+	return command.String()
+}
+
+//BuildKubectlCommandArgs build kubectl command to get logs
+func BuildKubectlCommandArgs(kubeconfig string, namespace, podName, container string, tail int64, sinceSeconds time.Duration) []string {
```
petersutter, line 29: change name to `BuildLogCommandArgs`

andrei-panov: method renamed

pkg/cmd/logs.go
```diff
@@ -225,14 +225,15 @@ func showLogsFromKubectl(namespace, toMatch, container string) {
 	checkError(err)
 	for _, pod := range pods.Items {
 		if strings.Contains(pod.Name, toMatch) {
-			err := ExecCmd(nil, buildKubectlCommand(namespace, pod.Name, container), false)
+			output, err := ExecCmdReturnOutput("kubectl", kubectl.BuildLogCommandArgs(KUBECONFIG, namespace, pod.Name, container, flags.tail, flags.sinceSeconds)...)
```
dansible, line 228: It is not always the case that the `KUBECONFIG` environment variable will be set. 

The usual flow is to run `gardenctl target ...` and export the output of that command if we want to work with `kubectl` directly but users can also run `gardenctl kubectl ...` without needing to export the `KUBECONFIG` var in their environment

dansible (changes requested):

pkg/cmd/logs.go
andrei-panov: The original implementation used it so I kept the same behavior.
```
func buildKubectlCommand(namespace, podName, container string) string {
	var command strings.Builder
	command.WriteString(fmt.Sprintf("kubectl logs --kubeconfig=%s %s%s -n %s ", KUBECONFIG, podName, container, namespace))
```

andrei-panov: Or you mean that I should handle this case?

andrei-panov:
Here I want to put some arguments about why another package is for good (subcmd/kubectl)
1) **It reduces the scope.**
Once we agreed to have all the functionality in this package than when you open the kubectl.go file you will be confident that you working only with this limited scope. You even allowed to introduce private methods not accessible outside the package. We have some *private* methods currently but it doesn't work since it's the same package everywhere.
2) **reduce spaghetti code**
As an example: KUBECONFIG variable setted up and readed from many different places. If you decide to collect *all* the logs from different places (garden, seed, shoot) in parallel, it becomes impossible because it's out of control when the variable will be updated. Get one by one is the only option for now.
3) **smaller package is more testable**
Smaller package scope gives you more confidence during refactoring.
4) **extension functions**
It's not directly related to a separate package, but once we move all the functionality related to kubectl to a dedicated package we can refactor the functions to extension functions over some structure. It will reduce the number of arguments which we pass here and there... Same way like go-client does.
As an example, we can have KUBECONFIG structure and then call `config.getLogs()` or `config.getNS()` and execute it against different kubeconfigs in parallel. It might be treated as syntax sugar but it's present in go-land so there no reason not to use it.

I believe that more or less the same arguments might be applied to all other subcommands which we call.
So please treat it as a proposal in general.

PS:
```
Running Suite: Kubectl Suite
============================
Ran 2 of 2 Specs in 0.000 seconds
SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
coverage: 100.0% of statements
```
you see... 100% for the whole suite :) It's not the goal... but who knows... what will convince you :)

pkg/cmd/logs.go
andrei-panov: As we discussed KUBECONFIG is not an environment variable and nothing should be changed in this sense.

dansible:
Hi @andrei-panov, I totally agree regarding reducing the scope of changes and making it easier to test and extend features. The issue is that these changes are outside of the scope of this PR/issue, which is for sanitizing the inputs to the logs subcommand. The repackaging work is a big refactor that should be tracked in its own issue and handled with smaller PRs that focus on one subcommand at a time. If this is something that you'd like to work on, I would encourage you to open an issue to track this either just for the `kubectl` subcommand or a larger umbrella issue that describes the proposed outcome for all packages.

Regarding the `logs all` target, the PR for that is here: https://github.com/gardener/gardenctl/pull/437

andrei-panov:
@dansible Please review it

dansible (approved):

PR #431, opened 2020-11-06
Make logPodSeed use the current context

### LINKED ISSUE #388: Make logPodSeed use the current context
We want to make `logPodSeed` function to use the current context if there's one available.

jguipi:
- Remove useless flag on the `logPodShoot` function
- Allow the `gardenctl logs vpn-seed` to use the current context 

Fixes #388 

- none

**Release note**:
```improvement operator
- Remove useless flag on the `logPodShoot` function
- Allow the `gardenctl logs vpn-seed` to use the current context 
```

### DIFF
--- a/pkg/cmd/logs.go
+++ b/pkg/cmd/logs.go
@@ -140,7 +140,11 @@ func runCommand(args []string) {
 	case "addon-manager":
 		logsAddonManager()
 	case "vpn-seed":
-		logsVpnSeed(args[1])
+		if len(args) == 2 {
+			logsVpnSeed(args[1])
+		} else {
+			logsVpnSeed(emptyString)
+		}
 	case "vpn-shoot":
 		logsVpnShoot()
 	case "machine-controller-manager":
@@ -473,7 +477,6 @@ func logPodShoot(toMatch, namespace string, container string) {
 	Client, err = clientToTarget(TargetKindShoot)
 	checkError(err)
 	if container != emptyString {
-		container = " -c " + container
 		showLogsFromKubectl(namespace, toMatch, container)
 	} else {
 		showLogsFromKubectl(namespace, toMatch, emptyString)
@@ -573,9 +576,14 @@ func saveLogsControllerManager() {
 }
 
 // logsVpnSeed prints the logfile of the vpn-seed container
-func logsVpnSeed(shootName string) {
+func logsVpnSeed(shootTechnicalID string) {
 	fmt.Println("-----------------------Kube-Apiserver")
-	logPodSeed("kube-apiserver", shootName, "vpn-seed")
+	if shootTechnicalID == emptyString {
+		shootTechnicalID = getFromTargetInfo("shootTechnicalID")
+		logPodSeed("kube-apiserver", shootTechnicalID, "vpn-seed")
+	} else {
+		logPodSeed("kube-apiserver", shootTechnicalID, "vpn-seed")
+	}
 }
 
 // logsEtcdOpertor prints the logfile of the etcd-operator


pkg/cmd/logs.go
```diff
@@ -382,9 +385,15 @@ func logsControllerManager() {
 }
 
 // logsVpnSeed prints the logfile of the vpn-seed container
-func logsVpnSeed(shootName string) {
+func logsVpnSeed(shootTechnicalID string) {
 	fmt.Println("-----------------------Kube-Apiserver")
-	logPodSeed("kube-apiserver", shootName, "vpn-seed")
+	if shootTechnicalID == emptyString {
+		output, err := ExecCmdReturnOutput("bash", "-c", "kubectl config current-context")
```
dansible, line 391: Here if the user already has a context defined we should not need to run the kubectl command again

dansible (changes requested):

pkg/cmd/logs.go
jguipi: It's actually through the command that I'm getting the context. Is there another way to get it ?

dansible: ah i see! Nvm my previous message in that case 😅 

I would just say to invoke `kubectl` directly without relying bash (something like `ExecCmdReturnOutput("kubectl", "config", "current-context"`)

jguipi: Done

pkg/cmd/logs.go
```diff
@@ -382,9 +385,15 @@ func logsControllerManager() {
...
+func logsVpnSeed(shootTechnicalID string) {
 	fmt.Println("-----------------------Kube-Apiserver")
-	logPodSeed("kube-apiserver", shootName, "vpn-seed")
+	if shootTechnicalID == emptyString {
+		output, err := ExecCmdReturnOutput("kubectl", "config", "view", "--minify", "--output", "jsonpath={..namespace}")
+		checkError(err)
+		logPodSeed("kube-apiserver", output, "vpn-seed")
+	} else {
+		logPodSeed("kube-apiserver", shootTechnicalID, "vpn-seed")
+	}
```
petersutter, line 396:
```suggestion
	var err error
	if shootTechnicalID == emptyString {
		shootTechnicalID, err = ExecCmdReturnOutput("kubectl", "config", "view", "--minify", "--output", "jsonpath={..namespace}")
		checkError(err)
	}
	logPodSeed("kube-apiserver", shootTechnicalID, "vpn-seed")
```

however, instead of reading the namespace from the current context, can't you make use of `GetFromTargetInfo(targetReader, "shootTechnicalID")` to get the technicalID?

jguipi: @petersutter Great suggestion but from some quick test seems like the `GetFromTargetInfo()` method only work on the shoot and return the following error when executed on the seed 
image

the followiing code seems to be the good way to go about it for now :

petersutter: ah ok, yes this won't work then

petersutter: the suggested change is still valid tho (independent of GetFromTargetInfo)

tedteng: sorry I just curious how can I reproduce it? I may need to check whether is any defect in the 'getFromTargetInfo' method and fix it.

however when I pull this PR and test it. it seems good. maybe I miss something in your scenario.  Please let me know, thanks 


image

image

image

jguipi: @tedteng You were right I was setting my environment the wrong way. It's working now ! Thx

dansible (approved):

Issue #385, opened 2020-10-15
Remove OIDC warning message


dansible:
The warning message we receive from gardenctl when using a kubeconfig with an `exec` arg (for OIDC) can be confusing and is not needed. GKE/EKS, and many other K8s platforms rely on similar kubeconfigs using an `exec` argument and do not display such warnings. Can we remove the OIDC warning message when using a kubeconfig with an `exec` arg?

neo-liang-sap:
Hi @dansible , this warning message is required by dev esp from security perspective and introduced in PR https://github.com/gardener/gardenctl/pull/221, you can see more discussion in https://github.com/gardener/gardenctl/pull/221#issuecomment-658239370

Some background:
1) original gardenctl doesn't allow kubeconfig contains `exec`
2) with OIDC enabled, we have to accept `exec` in kubeconfigs used by gardenctl
3) as a warning message / from security perspective this warning message is introduced, the aim is to warn use that they are at their own risk to use kubeconfigs which contains exec (might be macilious code)

Thanks.
-Neo

petersutter:
see also https://banzaicloud.com/blog/kubeconfig-security/

dansible:
Yes, I understand the issue and the background for it, however, constantly producing warnings whenever using an OIDC kubeconfig is distracting and can also be confusing. When we look at other platforms like GKE or EKS, they both use executable kubeconfigs but don't produce any such warnings. `kubectl` itself also doesn't provide any warnings when using these kubeconfigs. It should be the responsibility of the user to verify that their kubeconfig is valid. I don't think `gardenctl` needs to warn them every time a kubeconfig with an exec key is used, especially if it's for a valid setup like OIDC using `kube-login`

tedteng:
I have one feasible solution(only on Mac),  What about adding one method at the begging of `ValidateClientConfig ` display the full Warning message when the function invoked, then bring an option to the user asking whether need to display the Warning message all the time, or dismiss 3 day

if Yes create one tmp file under /tmp folder on local for checking. Warning message will skips if tmp file exists. 

By default, files that aren't accessed in three days are deleted from /tmp in Mac. https://developer.apple.com/forums/thread/71382

In this way, we can remove the OIDC warning message intermittently and also noticed msg with the user regularly as well. 
@dansible @petersutter  WDYT?

[...]

PR #439, opened 2020-11-12
option enable/disable oidc msg

tedteng:
option for enable/disable oidc warning msg
Fixes https://github.com/gardener/gardenctl/issues/385

**Release note**:

```improvement operator
enable/disable oidc warning msg
```


petersutter:
I'd rather:
Check if the (targeted) kubeconfig is considered to be safe.
If not and the user hasn't confirmed yet that it's trusted, warn the user and ask if he wants to trust this kubeconfig and store this information (hash for trusted kubeconfig)
Usage of the kubeconfig should be prevented as long as the user hasn't confirmed it to be trusted, wdyt?
The user should be informed/warned at least once for the respective kubeconfig (or there should be a summary of the kubeconfigs that could be potentially unsafe)

If I understood your code correctly, you just ask the user if the warnings should be disabled, without even checking if there is a potentially unsafe kubeconfig and without warning the user once. If there is no kubeconfig that is considered to be unsafe, there is no need to ask and no need to display a warning of course.

- Apart from that you are talking about a "OIDC warning" but it's rather a warning about a potentially unsafe kubeconfig
- Extract the code, that asks the user for confirmation, to a function
- Why checking only the first garden cluster configuration: [`switch gardenConfig.GardenClusters[0].KubeConfigTrusteds`](https://github.com/gardener/gardenctl/pull/439/files#diff-8a970d51dea43e926c0b913c26887464b362c5f723fb1d740a5a07d428434c07R255)
- rename `KubeConfigTrusteds` to something like `TrustedKubeConfigMd5` or `TrustedKubeConfigHash` and store the value with the hashing algorithm used, e.g. `{md5}3c3ec367dd9797153e4610b18a896a7b`

tedteng:
> The user should be informed/warned at least once for the respective kubeconfig (or there should be a summary of the kubeconfigs that could be potentially unsafe)

sure, in my first version on my local I used to have this. I will add a summary of the kubeconfigs  that could be potentially unsafe

tedteng:
> * Why checking only the first garden cluster configuration: [`switch gardenConfig.GardenClusters[0].KubeConfigTrusteds`](https://github.com/gardener/gardenctl/pull/439/files#diff-8a970d51dea43e926c0b913c26887464b362c5f723fb1d740a5a07d428434c07R255)

In my code logic,  there not any value `TrustedKubeConfigMd5` When people set up their local environment (`setup-kubeconfigs.sh`) 

```
cat ~/.garden/config
gardenClusters:
- name: dev
  kubeConfig: /Users/i333878/gardener_kubeconfig/dev/garden-gke/kubeconfig
  dashboardUrl: https://dashboard.garden.dev.k8s.ondemand.com
```
 So I only need to check the first one [0] then `kubeConfigMd5sumInit` to generated md5 for all garden landscape. make like 
```
gardenClusters:
- name: dev
  kubeConfig: /Users/i333878/gardener_kubeconfig/dev/garden-gke/kubeconfig
  dashboardUrl: https://dashboard.garden.dev.k8s.ondemand.com
  TrustedKubeConfigMd5: de7d1f4f16ed6233247a9129e8def31c
```
if people choose `n` 
```
gardenClusters:
- name: dev
  kubeConfig: /Users/i333878/gardener_kubeconfig/dev/garden-gke/kubeconfig
  dashboardUrl: https://dashboard.garden.dev.k8s.ondemand.com
  TrustedKubeConfigMd5: none

**Comment by tedteng** [comment]
> If I understood your code correctly, you just ask the user if the warnings should be disabled, without even checking if there is a potentially unsafe kubeconfig and without warning the user once. If there is no kubeconfig that is considered to be unsafe, there is no need to ask and no need to display a warning of course.

with the latest code, I change like this. 

- all garden landscape kubeconfig verification and display msg at the first time when user use 
- then generate md5 for all garden landscape kubeconfig 
- always verify md5 , display confirm msg only when md5 not match. 

more testing result
1) target at the first time verification and display msg
2) target at the second time without msg
3) modify source kubeconfig file bring the confirm msg 
image


image

**Comment by petersutter** [comment]
I just tested it and I saw the warning message but I was never asked if I trust it, probably because of [this line] (https://github.com/gardener/gardenctl/pull/439/files#diff-8a970d51dea43e926c0b913c26887464b362c5f723fb1d740a5a07d428434c07R272)
```go
kubeConfigMd5sumInit("y", gardenConfig)
```
I also had yaml comments in my garden config which are now gone. This probably happened once you updated the config with the hashes.

when I said 
> Extract the code, that asks the user for confirmation, to a function
I was thinking about a function with such a signature (that is returning a bool value) `func askForConfirmation() bool` where the yes / no input is handled.

/invite @gardener/gardenctl-maintainers please also review this PR

tedteng:
>Extract the code, that asks the user for confirmation, to a function
I was thinking about a function with such a signature (that is returning a bool value) func askForConfirmation() bool where the yes / no input is handled.

done 

>I just tested it and I saw the warning message but I was never asked if I trust it, probably because of [this line]

done, confirm first before force hash

tedteng:
> I also had yaml comments in my garden config which are now gone. This probably happened once you updated the config with the hashes.

after testing, It seems comments will remove it. not sure this is handle by golang write itself or logic from somewhere. I will take a look at whether possible keep the comments.

tedteng:
@petersutter  The comments is seems not support yaml v2 , and I havn't achieve it base on yaml v3 yet.

from the example, if I understand correctly  it need to change all element in `GardenClusterMeta` struct to yaml.Node type https://godoc.org/gopkg.in/yaml.v3#Node.  then we can keep comments
as consequence which affect more go files need refactor which like `cannot convert garden.Name (type "gopkg.in/yaml.v3".Node) to type string` . I don't know whether I am not the right approach. Let me know WDYT? 


For example:

var person struct {
        Name    string
        Address yaml.Node
}
err := yaml.Unmarshal(data, &person)

petersutter:
OK, it's not a top priority for me that the comments do not survive

tedteng:
revoke the last commit and close the PR will create the new one.

PR #441, opened 2020-11-13
enable loki-0 for cluster

andrei-panov:
purpose development is sufficient for enabling loki on cluster

enable loki-0 on cluster for testing

### DIFF
--- a/.github/gardener_clusters/gctl-gcp.yaml
+++ b/.github/gardener_clusters/gctl-gcp.yaml
@@ -77,7 +77,7 @@ spec:
           - europe-west1-d
         systemComponents:
           allow: true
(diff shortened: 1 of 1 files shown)


.github/gardener_clusters/gctl-gcp.yaml
```diff
@@ -77,7 +77,7 @@ spec:
           - europe-west1-d
         systemComponents:
           allow: true
-  purpose: testing
+  purpose: development
```
dansible, line 80: You will need `production` purpose to enable the monitoring stack

dansible (changes requested):

dansible (approved):

PR #445, opened 2020-11-16
cleanup function will return the code which ssh returns

### LINKED ISSUE #444: gardenctl ssh <openstack_node> fails with `Operation timed out`
**Describe the bug**
`gardenctl ssh <openstack_node>` fails with `Cmd was unsuccessful` (using `gardenctl/gctl-os` cluster)

**To Reproduce**
Using 52235f2f with a test cluster gives:

```
$ go run cmd/gardenctl/main.go ssh shoot--gardenctl--gctl-os-worker-lx07a-z1-f46b4-pbp2b exit

Warning:
Be aware that you are entering an untrusted environment!
Do not enter credentials or sensitive data within the ssh session that cluster owners should not have access to.

Downloaded id_rsa key
Check Public IP
(1/5) Getting the external network for creating FIP
The external network ID is f55c8e3d-c798-4c6c-9f7e-eeb6600f6aed
(2/5) Creating floating IP from external network
The floating IP created is XXXX
(3/5) Add floating IP to openstack server node
(4/5) Establishing SSH connection

ssh: connect to host XXXX port 22: Operation timed out
exit status 255

(5/5) Cleanup
De-associate server with floating ip
Delete the floating IP


$ echo $?
0
```

Unfortunately, this also doesn't return an error because the cleanup operation succeeded. It would be great to also expose the error code to the end user (even if the cleanup was successful)

 - Version/Commit 52235f2f

neo-liang-sap:
cleanup function will return the code which ssh returns
Fixes https://github.com/gardener/gardenctl/issues/444 and https://github.com/gardener/gardenctl/issues/448

**Release note**:
```improvement operator
cleanup function will return the code which ssh returns
```

### DIFF
--- a/pkg/cmd/ssh_openstack.go
+++ b/pkg/cmd/ssh_openstack.go
@@ -71,6 +71,12 @@ func sshToOpenstackNode(nodeName, path, user, pathSSKeypair string, sshPublicKey
 
 	defer a.cleanUpOpenstack(&sshStatus)
 
+	err = CheckIPPortReachable(a.FIP, "22")
+
+	if err != nil {
+		sshStatus = 1
+	}
+
 	node := user + "@" + a.FIP
 	fmt.Println("(4/5) Establishing SSH connection")
 	fmt.Println("")
--- a/pkg/cmd/ssh_openstack.go
+++ b/pkg/cmd/ssh_openstack.go
@@ -69,6 +69,8 @@ func sshToOpenstackNode(nodeName, path, user, pathSSKeypair string, sshPublicKey
 	operate("openstack", "server add floating ip "+a.InstanceID+" "+a.FIP)
 	time.Sleep(5000)
 
+	defer a.cleanUpOpenstack(&sshStatus)
+
 	node := user + "@" + a.FIP
 	fmt.Println("(4/5) Establishing SSH connection")
 	fmt.Println("")
@@ -92,11 +94,10 @@ func sshToOpenstackNode(nodeName, path, user, pathSSKeypair string, sshPublicKey
 			sshStatus = exitError.ExitCode()
 		}
 	}
-	defer a.cleanUpOpenstack(sshStatus)
 }
 
 //cleanUpOpenstack clean the resource added to ssh to openstack node
-func (a *OpenstackInstanceAttribute) cleanUpOpenstack(exitStatus int) {
+func (a *OpenstackInstanceAttribute) cleanUpOpenstack(exitStatus *int) {
 	fmt.Println("")
 	fmt.Println("(5/5) Cleanup")
 
@@ -105,5 +106,5 @@ func (a *OpenstackInstanceAttribute) cleanUpOpenstack(exitStatus int) {
 
 	fmt.Println("Delete the floating IP")
 	operate("openstack", "floating ip delete "+a.FIP)
-	os.Exit(exitStatus)
+	os.Exit(*exitStatus)
 }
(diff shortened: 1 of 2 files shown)


neo-liang-sap:
Hi @dansible ,
I tested in following two scenario
1) postive path - ssh returns OK and the exit status code is 0
2) negative path - i fake ssh username to "neo" and ssh failed with exit code 255, then overall exit status is 255

BTW the openstack infra itself is not very stable - in your case i saw you were unable to ssh to gctl-os shoot but today i can ssh without error, this is the same when i was doing gardenctl ssh openstack_node - sometimes the ssh failed for unknown reason...i didn't dig out why, retry will success....

image
image

dansible (approved):

pkg/cmd/ssh_openstack.go
```diff
@@ -89,11 +88,15 @@ func sshToOpenstackNode(nodeName, path, user, pathSSKeypair string, sshPublicKey
 	cmd.Stderr = os.Stderr
 	if err := cmd.Run(); err != nil {
 		fmt.Println(err)
+		if exitError, ok := err.(*exec.ExitError); ok {
+			sshStatus = exitError.ExitCode()
+		}
 	}
+	defer a.cleanUpOpenstack(sshStatus)
```
petersutter, line 95: what's the purpose of having `defer` as last statement of the function? I would refrain from moving the deferred cleanup call to the bottom.

neo-liang-sap: during my test, if i move the `defer a.cleanUpOpenstack(sshStatus)` to above lines of value assignment of `sshStatus `, it won't get correct values.
Say if i keep `defer a.cleanUpOpenstack(sshStatus)`  to original line, when some error occurs in ssh, `sshStatus` will be assigned with value 1 (for example) but in `defer a.cleanUpOpenstack(sshStatus)` the value of `sshStatus` is still `0`
do you have a better idea how i can get `sshStatus` correctly passed to `cleanUpOpenstack` meanwhile i can keep the original line for the defer function? thanks!

petersutter: if you want to achieve this you need to pass a pointer

neo-liang-sap: thanks! i will have a try

neo-liang-sap: Hi @petersutter , i've modified the code using *int to pass values, thanks for your suggestion!
-Neo

neo-liang-sap:
@dansible , per our discussion in planning mtg this week, i implement a function to check whether IP and port are reachable (https://github.com/gardener/gardenctl/issues/448) , i implement this feature in this PR as this function will impact the overall exit status so i append this feature together in this PR. Please review it when you have time, thanks!
-Neo

petersutter (changes requested):

neo-liang-sap:
thanks @petersutter , i used go native method to check whether the ip and port reachable

pkg/cmd/utils.go
```diff
@@ -348,3 +350,18 @@ func PrintoutObject(objectToPrint interface{}, writer io.Writer, outputFormat st
 	}
 	return nil
 }
+
+//CheckIPPortReachable check whether IP with port is reachable with 1 min
+func CheckIPPortReachable(ip string, port string) error {
+	timeout := time.Second * 60
+	conn, err := net.DialTimeout("tcp", net.JoinHostPort(ip, port), timeout)
```
petersutter, line 357: and this works without having to dial/try multiple times within a certain amount of time?

neo-liang-sap: Yes, in my test it tries within 1 min, as I set in the timeout

petersutter (approved):
/lgtm

dansible (approved):

PR #465, opened 2020-11-25
add gardenctl hibernate and gardenctl wakeup to hibernate/wakeup shoot cluster

### LINKED ISSUE #450: implement `gardenctl wakeup` to wake up a shoot cluster and `gardenctl hibernate` to hibernate a shoot cluster after shoot targeted
as titled.
After shoot is targeted,

If the shoot is hibernated, `gardenctl wakeup` will wakeup the shoot, wait until shoot waked up.
If the shoot is not hibernated, print out msg indicating the shoot status.

If the shoot is not hibernated, `gardenctl hibernate` will hibernate the shoot, wait until shoot hibernated.
If the shoot is hibernated, print out msg indicating the shoot status.
/assign

neo-liang-sap:
add gardenctl hibernate and gardenctl wakeup to hibernate/wakeup shoot cluster
Fixes https://github.com/gardener/gardenctl/issues/450

**Release note**:
```improvement operator
after targeting a shoot, you could run "gardenctl hibernate" / "gardenctl wakeup" to hibernate / wakeup a shoot
```


neo-liang-sap:
if no objection i will merge this PR by EOB today

petersutter (changes requested):
please do not merge unreviewed PRs.

After scrolling quickly over the PR:
- Why do you fetch all shoots across all namespaces if you just want to get one?
- There is at least one typo: `descirption` (`Last operation descirption is`
- Instead of polling the shoot, why don't you watch for changes?
- I would not compare the Description, this would be a fragile solution `newShoot.Status.LastOperation.Description == "Shoot cluster state has been successfully reconciled."`. The description may be changed any time by the gardener

neo-liang-sap:
Hi @petersutter , thanks for your review and comments.
- i've modified shoot namespace while doing list of shoots, now it's `shootList, err := gardenClientset.CoreV1beta1().Shoots(shoot.GetNamespace()).List(metav1.ListOptions{})`

- i've removed description compare, so the typo is also removed

- for `watch for changes`, i'm not understanding very clearly. Do you mean move the logic of getting newShoot outside of for loop? in my test, the newShoot object is immutable object,  this object will stay the same without reflecting new status change. I'm not sure whether i have expressed myself clearly, if i put logic of getting newShoot outside of for loop the result would be sth like this
image
The progress won't change.

So 'each time pulling newShoot object' is the simplest way i can come up with, in my test i see the performance is OK , as `gardenctl hibernate/wakeup` checks shoot status at 1 min interval, this waiting interval takes most of the running time.

Do you have a better way for getting & comparing the newShoot status? it would be much appreciate if you can propose some  code snippet for me to reference, thanks!
-Neo

petersutter:
Hi @neo-liang-sap ,

- In general, if you are only interested in one shoot and you have the namespace and name of the shoot, as you do in this case, please use `get` instead of `list`. But if you watch for changes you anyhow need to do it differently
- Take this as an example on how to implement a watch a resource
  - watch with 10 seconds timeout https://github.com/gardener/terminal-controller-manager/blob/master/controllers/terminal_controller.go#L1255-L1258
  - https://github.com/gardener/terminal-controller-manager/blob/master/controllers/terminal_controller.go#L833-L883

Implementing a watch is not required of course, but definitely more efficient and instant

neo-liang-sap:
Hi @petersutter , thanks, i've changed code to use `get` to get shoot.
For `watch` example you provided i took look at them and it seems would takes much effort to implement similar in `hibernate`/`wakeup` function, as this function doesn't have strict performance requirement i wonder is it ok to use current implementation (get the shoot object at 1 min interval and check the status) as a easy solution?
Thanks.
-Neo

pkg/cmd/utils.go
```diff
@@ -370,3 +373,56 @@ func CheckIPPortReachable(ip string, port string) error {
...
 }
+
+//setHibernation hibernate a shoot or wake up a shoot
+func setHibernation(shoot *gardencorev1beta1.Shoot, hibernated bool) {
+	if shoot.Spec.Hibernation != nil {
+		shoot.Spec.Hibernation.Enabled = &hibernated
+	}
+	shoot.Spec.Hibernation = &gardencorev1beta1.Hibernation{
+		Enabled: &hibernated,
+	}
```
vpnachev, line 384: This will overwrite hibernation schedule.

neo-liang-sap:
per customer requirement i'm reopening this issue https://github.com/gardener/gardenctl/issues/450#issuecomment-838292497
will continue when available
/reopen

PR #504, opened 2021-02-25
Gactl cluster

jguipi:
Update gardener cluster template

Fixes  none

- none
**Release note**:
```other operator
Update gardener cluster template
```

### DIFF
--- a/.github/gardener_clusters/gctl-ali.yaml
+++ b/.github/gardener_clusters/gctl-ali.yaml
@@ -39,7 +39,7 @@ spec:
         cpu: 80m
         memory: 1Gi
         pid: 20k
-    version: 1.19.3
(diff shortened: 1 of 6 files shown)


neo-liang-sap:
@jguipi - this PR is regarding testing shoots right? do we still need this ? i mean if we need it i will proceed with merge, or close if we don't need it any more

jguipi:
> @jguipi - this PR is regarding testing shoots right? do we still need this ? i mean if we need it i will proceed with merge, or close if we don't need it any more

I'm not sure if we still need it, we were using those cluster for gardener testing. I think we can still merge it and keep gardener template up to date for future use.

neo-liang-sap:
proceed with merging per disucssion

